PR #53, opened 2023-09-20
Dev

patrickmohrmann:
basic setup of less and gulp for automatic css creation.
added lang files and fonts
added a few assets


assets/token/token.png
ChristopherSD, line 1: maybe rename to something like "token-frame-circular", so it's absolutely clear what this file is supposed to represent

assets/wallpaper/heroes-legends-and-foes.jpg
ChristopherSD, line 1: maybe add "-wallpaper" or something at end of file name, so it's clear what the file represents.

gulpfile.js
ChristopherSD, line 1: If we follow the DnD system, we should probably use `.mjs` file extensions, so its cleary we're using ES6.

The code is fine for now. We can have a look later if we add additional things like compiling and linting.

ChristopherSD (commented):
I would suggest we change the way the language files are constructed. Using the first letter as a grouping does not add additional information, but just more to write. Better group it by things like entity type, by application (e.g., `ED4E.ActorSheet.Characteristics.Dexterity`) or something of that sort.

ChristopherSD (approved):
Changes will be added later.
(This is the only time we do this :D )

PR #128, opened 2023-10-04
Patrick dev 02

patrickmohrmann:
test


module/sheets/actor-sheet.mjs
ChristopherSD, line 1: This needs to be arranged in the ES6 module directory structure. Do you want to do this together or should I do it separately and make an extra pull request for that?

Line 20: No trailing comma, please remove.

module/sheets/item-sheet.mjs
ChristopherSD, line 1: This needs to be arranged in the ES6 module directory structure. Do you want to do this together or should I do it separately and make an extra pull request for that?

earthdawn4e.mjs
ChristopherSD, line 1: The imports need to be arranged in the ES6 module directory structure. Do you want to do this together or should I do it separately and make an extra pull request for that?

Line 53: The function should not return anything, since the Promise of the `preloadHandlebarsTemplates` is ignored anyways.

module/helpers/handlebar-helpers.js
ChristopherSD, line 1: This needs to be arranged in the ES6 module directory structure. Do you want to do this together or should I do it separately and make an extra pull request for that?

The file should have the `*.mjs` extension.

This can be consolidated into a `utils.mjs` file.

module/helpers/templates.mjs
ChristopherSD, line 1: This needs to be arranged in the ES6 module directory structure. Do you want to do this together or should I do it separately and make an extra pull request for that?

This can be consolidated into a `utils.mjs` file.

The `loadTemplates` function also allows for object of templates as a parameter. This way you do not need to write the entire file name when using a partial in an hbs-file, but rather can use the shorter key. Here is a code snippet that creates the object automatically:
```javascript
const partials = [
    "systems/ed4e/templates/actor/character-partials/header-section.hbs",
    ...
]

const paths = {};
  for ( const path of partials ) {
    paths[path.replace(".hbs", ".html")] = path;
    paths[`ed4e.${path.split("/").pop().replace(".hbs", "")}`] = path;
  }

  return loadTemplates(paths);
```

ChristopherSD (changes requested):
See comments in files.

earthdawn4e.mjs
patrickmohrmann: fine if you do it, I can figure it out in the pull request

module/sheets/actor-sheet.mjs
patrickmohrmann: will be handled in future pull

module/sheets/item-sheet.mjs
patrickmohrmann: will be handled in future pull

patrickmohrmann (commented):
update

ChristopherSD (approved):

PR #290, opened 2023-12-06
Patrick clean up

patrickmohrmann:


module/config.mjs
ChristopherSD, line 1: Any reason the whitespace is different here than the rest of file?

patrickmohrmann: what whitespace?

module/data/actor/pc.mjs
patrickmohrmann, line 1: this fixed #292

module/config.mjs
ChristopherSD: okay, as long as we're doing it consistently across the file, either one is fine for me

ChristopherSD (approved):

PR #308, opened 2023-12-19
Patrick dark mode

patrickmohrmann:


ChristopherSD (commented):
The entire attribute table should be written as a Handlebars {{#each}}. Makes for less copy-paste errors.

ChristopherSD (approved):
Streamlined a few parts in code. See recent commits.

Suggestion: rename property `globalBonuses` to just `modifier`. More concise and slightly more correct since it's not just bonuses, but also penalties.

If they are okay, fine to merge.

PR #326, opened 2023-12-26
fixed, attribute names

patrickmohrmann:
attribute names are now shown as localized label not abbreviation

#324


patrickmohrmann:
pushed in chris-roll branch

Issue #320, opened 2023-12-22
add a lock and refresh option for current karma and calculate Karma


patrickmohrmann:
calculate karma points for an actor of type character:

KarmaMaximum = karmaValue of Namegiver item * highest circle + unspend Attribute points.

add a lock option to enable writability of Karma values (current, max and step)

enable the refresh function to reset current karma to max.

patrickmohrmann:
done with PR #327

PR #327, opened 2023-12-27
#320 karma calculation and lock

patrickmohrmann:
the karma rituatl button is now refreshing current karma to maximum Karma


module/applications/actor/base-sheet.mjs
```diff
@@ -120,4 +123,11 @@ export default class ActorSheetEd extends ActorSheet {
     const item = this.actor.items.get( li.dataset.itemId );
     return item.sheet?.render( true );
   }
+
+  async _onKarmaRefresh ( ) {
```
ChristopherSD, line 127: Just like with the weightCalculation, this should an appropriate method in the actor document, e.g. `this.actor.refreshKarma();`.
The function in the actor document class should like this, using the update method to persist the changes:
```javascript
karmaRitual() {
    this.update( {"system.karma.value": this.system.karma.max} );
}
```

ChristopherSD (changes requested):
One change needed in base-sheet.mjs

module/applications/actor/base-sheet.mjs
patrickmohrmann: ok, moved it as I understood it

ChristopherSD: Did some minor cleanup in commit 455b8f8cc3aee330be00d33c1a2ed8c1c9dc426a. If that is fine, mark this as resolved and we can merge it.

patrickmohrmann (commented):
good

PR #360, opened 2024-01-05
248 visualize ability content on actor sheet

patrickmohrmann:
closes #248 

it is a future topic nesessary for ability rolls, but containing only html & css enhancements

PR #363, opened 2024-01-05
362 apply visualization for rollable items to physical items

patrickmohrmann:


ChristopherSD (commented):
Can it be that there are commits in this branch that are also present in #360 ?
If so, that is really confusing and will probably lead to quite some merging conflicts...

patrickmohrmann:
yes, it builds on #360

Issue #460, opened 2024-01-29
remove strain & modifier from abitrary roll from dice


patrickmohrmann:
the dice symbol triggering an arbitrary step shall not include the modifier nor the strain fields

PR #461, opened 2024-01-29
328 add actor token image to chat output

### LINKED ISSUE #328: add actor token image to chat output
Add a system setting:

- show token image in chat out put
- Global setting
- ~GMs shall have the option to insert a GM image~ (or maybe its possible to use the configured one. 
- For the GM rolls the image of  the token / Aktor caused the roll shall be taken

options

patrickmohrmann:
closes #328 
and #460


module/settings.mjs
```diff
@@ -129,6 +129,28 @@ export default function registerSystemSettings() {
...
+    /* -------------------------------------------------------------------------------- */
+    // Chat Avatar settings Header
+    game.settings.register( "ed4e", "chatAvatarHeader", {
+        name: "ED.Settings.Chat.chatAvatarHeader",
+        config: true,
+    } );
+    // Chat Avater Options
+    game.settings.register( "ed4e", "chatAvatar", {
+        name: "ED.Settings.Chat.chatAvatar",
+        hint: "ED.Settings.Chat.chatAvatarHint",
```
ChristopherSD, line 144: Maybe `scope: "client"`? So players also can decide which way they want it.
See comments in chat.mjs

module/hooks/chat.mjs
```diff
@@ -117,10 +117,21 @@ function triggerRollStep(argString) {
 
 function addUserPortrait( msg, jquery ) {
 
-  const avatar = msg.user.avatar;
+  const chatAvatarSetting = game.settings.get( "ed4e", "chatAvatar" );
+  let avatar;
+  if ( chatAvatarSetting === "configuration" && game.user.isGM ) {
```
ChristopherSD, line 122: Is the `isGM` necessary? If setting's scope is `client` every player can decide for themselves?

module/hooks/chat.mjs
```diff
@@ -117,10 +117,21 @@ function triggerRollStep(argString) {
...
+    avatar = msg.user.avatar;
+  } else if ( chatAvatarSetting === "selectedToken" ) {
+    if ( canvas.tokens.controlled[0] ) {
```
ChristopherSD, line 125: Ich glaube der Block in diesem `else if` kann so zusammengefasst werden:
```javascript
avatar = canvas.tokens.controlled[0]?.document.texture.src ?? msg.user.avatar;
```

module/hooks/chat.mjs
```diff
@@ -117,10 +117,21 @@ function triggerRollStep(argString) {
 
 function addUserPortrait( msg, jquery ) {
```
ChristopherSD, line 118: If the `isGM` wouldn't be necessary, then the whole block could be streamlined, I think, if so wanted:

```javascript
avatar = chatAvatarSetting === "selectedToken"
    ? canvas.tokens.controlled[0]?.document.texture.src
    : msg.user.avatar;
```

Possible since the assignment in the first `if` is the same as the nested `else`.

templates/global/roll-prompt.hbs
```diff
@@ -28,7 +28,7 @@
 
 
                     <!-- 30/20 -->
-                    {{#if ( uneq rollType CONFIG.ED4E.rollTypes.arbitrary) }}
+                    {{#if ( uneq options.rollType arbitrary) }}
```
ChristopherSD, line 31: If I roll from the char sheet I don't see the strain and modifier options anymore.
This should be just `uneq rollType arbitrary`.  The rollOptions are unpacked in the context data, so they can just be accessed by name, without qualifier in front of it.

ChristopherSD (changes requested):

templates/global/roll-prompt.hbs
patrickmohrmann: ok check. solved it

PR #470, opened 2024-02-07
393 roll arbitrary step from journal entry

patrickmohrmann:


ChristopherSD (approved):
I streamlined operations and cleaned up. Following DnD5e v3.
Please have a look if that is okay for you.

- EdRoll needs to be adapted to accept absolute values, currently ignored
- EdRoll needs to be adapted for base step to have flavour as well?
- enrichers can be added as needed

PR #519, opened 2024-03-03
Update enrichers.mjs

patrickmohrmann:
bug fixed by checking if modifier extists.


ChristopherSD (approved):
Reviewed together in weekly

PR #608, opened 2024-03-12
tour & System documentation

patrickmohrmann:


ChristopherSD (changes requested):
Text has to be checked, but that comes later with localization.
If possible, maybe make the Journal Application a bit smaller, or rather, move it to left, otherwise the chat input and the dice symbol are covered by it.

patrickmohrmann:
sadly not possible to move those things around as far as i found out

ChristopherSD (approved):
Okay for now. Later, the window/app will be closed in between steps, so the sidebar is fully visible.

Issue #354, opened 2024-01-03
chat output for ability tests


patrickmohrmann:
show:

- [x] image
- [x] name
- [x] ability
- [x] successes
- ~apply effect (this will be done with EAE)~

patrickmohrmann:
closed with #620

Issue #250, opened 2023-11-04
add difficulty calculation based on sheet info


patrickmohrmann:
**Description**
The Roll prompt shall be enhanced by containing the following additional options:
- [ ] Difficulty tab 
  - [ ] difficulty
  - [ ] group difficulty

the roll prep shall automatically fetch / update the difficulty based on the actor targets selection.

**Additional context**

Issue #232, opened 2023-11-01
Ability Roll


patrickmohrmann:
## Reason

by clicking the dice symbol in front of an ability, a roll prompt shall be opened and a roll be triggered.
- [x] talent roll
- [x] skill roll
- [x] devotion roll
- [x] knack roll

some devotions require a devotion point to be used, those shall already set the required devotions to 1 and will throw an error if no devotions was used.

all abilities shall hand over a standard difficulty if one is available in the item details.

~some abilities have a certain range, those shall be checked against the current Target.~ (range will be encountered in token handling)

## Technical Details

patrickmohrmann:
closed with #620

PR #620, opened 2024-03-16
247 trigger ability test new

### LINKED ISSUE #247: Trigger Ability test
**Description**
ability tests shall be triggered by clicking on the dice symbol, which appears on a mouse over on the respective ability.

all abilitys shall have the rollType = ability

**technical details**
shall work similar for:
- talents
- skills
- devotions
- knacks

patrickmohrmann:
fixes #247 , #250 & #354 & #232


module/applications/global/document-creation.mjs
```diff
@@ -35,8 +35,8 @@ export class DocumentCreateDialog extends FormApplication {
       closeOnSubmit: false,
       submitOnChange: true,
       submitOnClose: false,
-      height: "auto",
-      width: "auto",
+      height: 900,
+      width: 800,
```
ChristopherSD, line 39: :D not really part of ability rolls? ;P

module/data/item/templates/ability.mjs
```diff
@@ -58,6 +58,28 @@ export default class AbilityTemplate extends SystemDataModel {
                 integer: true,
                 label: "ED.Item.Ability.strain"
             } ),
+            difficulty: new foundry.data.fields.SchemaField( {
```
ChristopherSD, line 61: Are there any other (edge) cases? E.g. like for spells where it's either a fixed difficulty OR the defense of the target whichever is higher/lower etc?
Also, how do we know which field to take if some reason more than one has a value?
We might need to rethink this...

module/data/item/templates/none-namegiver-power.mjs
```diff
@@ -43,6 +43,28 @@ export default class NoneNamegiverPowerData extends SystemDataModel {
                 integer: true,
                 label: "ED.Item.Power.damageStep"
             } ),
+            difficulty: new foundry.data.fields.SchemaField( {
```
ChristopherSD, line 46: Mhh, I forgot: why isn't this a subclass of `Ability` if it has all the same fields?

module/config.mjs
```diff
@@ -17,17 +17,18 @@ _______________________________`;
  */
 ED4E.typeGroups = {
 	Item: {
-		Equipment: ['armor', 'equipment', 'shield', 'weapon'],
-		Abilities: ['attack', 'devotion', 'knack', 'maneuver', 'power', 'skill', 'specialAbility', 'talent'],
+		Equipment: 	['armor', 'equipment', 'shield', 'weapon'],
```
ChristopherSD, line 20: not really ability test ;P

module/data/item/talent.mjs
```diff
@@ -11,7 +11,14 @@ export default class TalentData extends AbilityTemplate.mixin(
     /** @inheritDoc */
     static defineSchema() {
         return this.mergeSchema( super.defineSchema(), {
-            
+            talentType: new foundry.data.fields.StringField( {
```
ChristopherSD, line 14: Can we find a more descriptive name for that? Maybe something like "source", "sourceType", "learnedThrough" or something like that?

module/data/item/talent.mjs
```diff
@@ -11,7 +11,14 @@ export default class TalentData extends AbilityTemplate.mixin(
...
+                required: true,
+                nullable: false,
+                blank: false,
+                initial: "Discipline",
```
ChristopherSD, line 18: This must be lower case, as the keys in the config lookup object.

templates/item/item-partials/item-details/details/item-details-talent.hbs
```diff
@@ -1,3 +1,23 @@
 <div>
     <h3>{{localize "ED.Item.General.details"}}</h3>
-</div>
\ No newline at end of file
+</div>
+    {{!-- Target Difficulty --}}
```
ChristopherSD, line 4: If they are actually mutually exclusive the other two should be hidden when one has a value. But depends on how we implement the model.

templates/item/item-partials/item-details/descriptions/item-description-talent.hbs
```diff
@@ -33,6 +33,13 @@
             {{selectOptions config.tier selected=item.system.tier localize=true }}
         </select>
     </div>
+    {{!-- Tier --}}
+    <div class="item-data__input--line">
+        <label class="label__left">{{localize "ED.Tier.tier"}}:</label>
```
ChristopherSD, line 38: Wrong label. Should be something like "Source" not "Tier"

templates/item/item-partials/item-details/descriptions/item-description-devotion.hbs
```diff
@@ -36,7 +36,7 @@
     {{!-- Devotion Required --}}
     <div class="item-data__input--line">
         <label class="label__left">{{localize "ED.Item.Devotion.devotionRequired"}}:</label>
-        <input class="item-data__checkbox--line" type="checkbox" name="system.devotionRequired">
+        <input class="item-data__checkbox--line" type="checkbox" name="system.devotionRequired" {{checked item.system.devotionRequired}}>
```
ChristopherSD, line 39: not ability test ;P

templates/item/item-partials/item-details/details/item-details-attack.hbs
```diff
@@ -1,3 +1,23 @@
 <div>
     <h3>{{localize "ED.Item.General.details"}}</h3>
-</div>
\ No newline at end of file
+</div>
+    {{!-- Target Difficulty --}}
```
ChristopherSD, line 4: This should definitely a handlebar partial, as often as it's used.

module/handlebar-helpers.mjs
```diff
@@ -56,4 +56,15 @@ export default function registerHandlebarHelpers() {
   Handlebars.registerHelper( 'ed-linkForUuid', linkForUuid );
 
   Handlebars.registerHelper( 'ed-diceFormulaForStep', getDice );
+
+  Handlebars.registerHelper( "getTalentType", ( talents, type ) => {
+    return talents.filter( ( talent ) => talent.system.talentType === type );
+  } );  
+
+  Handlebars.registerHelper( "getKnackName", ( talent, knacks ) => {
```
ChristopherSD, line 64: StringCompare not good. That's why we use the ForeignDocument, or the UUIDFields.

ChristopherSD (commented):
There are a few points that we should still discuss during stand-up, since these are decision that have a heavy influence on some workflows.

Additionally, imo we should expand the datamodel to have a KnackTemplate, so all the different knack types can inherit from that.

module/data/other/roll-options.mjs
```diff
@@ -250,6 +250,13 @@ export default class EdRollOptions extends foundry.abstract.DataModel {
           step: 1,
           integer: true,
         } ),
+        pointsRequired: new foundry.data.fields.BooleanField( {
```
ChristopherSD, line 253: Why do we need that? The roll doesn't care if its required or not. That check should be done before rolling.

module/documents/actor.mjs
```diff
@@ -165,12 +246,16 @@ export default class ActorEd extends Actor {
...
-    this.takeDamage( roll.edRollOptions.strain, "standard" );
+    this.takeDamage( roll.options.strain.total, "standard", undefined, true );
     if (
-        !this.#useResource( 'karma', roll.edRollOptions.karma.pointsUsed )
-        || !this.#useResource( 'devotion', roll.edRollOptions.devotion.pointsUsed )
+        !this.#useResource( 'karma', roll.options.karma.pointsUsed )
+        || !this.#useResource( 'devotion', roll.options.devotion.pointsUsed )
     ) {
-      ui.notifications.warn("Localize: Not enough karma or devotion. Used all that was available.");
+      ui.notifications.warn( "Localize: Not enough karma or devotion. Used all that was available." );
```
ChristopherSD, line 254: These checks should be done before prompting the roll. It's pretty frustrating to only know something isn't possible after you prepared everything for it.
Maybe we need something like `#preProcessRoll` method

module/documents/actor.mjs
```diff
@@ -165,12 +246,16 @@ export default class ActorEd extends Actor {
       return;
     }
     // Check if this uses karma or strain at all
-    this.takeDamage( roll.edRollOptions.strain, "standard" );
+    this.takeDamage( roll.options.strain.total, "standard", undefined, true );
```
ChristopherSD, line 249: Accessing fields of classes from outside the class should be strictly controlled. So maybe we should add a Getter to the roll calls for that, like `get totalStrain()`

module/documents/actor.mjs
```diff
@@ -118,6 +118,86 @@ export default class ActorEd extends Actor {
     this.#processRoll( roll );
   }
 
+  /**
+   * @description                 Roll an Ability. use {@link RollPrompt} for further input data.
+   * @param {object} ability      ability.
```
ChristopherSD, line 123: Is the type `object` or `ItemEd`?

module/documents/actor.mjs
```diff
@@ -118,6 +118,86 @@ export default class ActorEd extends Actor {
...
+   * @param {object} options      Any additional options for the {@link EdRoll}.
```
ChristopherSD, line 124: The type should be `EdRollOptions` then.

module/documents/actor.mjs
```diff
@@ -118,6 +118,86 @@ export default class ActorEd extends Actor {
...
+   */
+  async rollAbility( ability, options = {} ) {
+    const attributeStep = this.system.attributes[ability.system.attribute].step;
+    const abilityFinalStep = attributeStep + ability.system.level;
+    const difficulty = await this.getTarget( ability );
+    const edRollOptions = new EdRollOptions( {
+      rollType: "action",
```
ChristopherSD, line 131: Depending on how we implement the combat workflow, this could also be damage (for things like Claw Shape, Body Control, etc.).
See Review of #636.

module/documents/actor.mjs
```diff
@@ -118,6 +118,86 @@ export default class ActorEd extends Actor {
...
+      target: { base: difficulty },
+      karma: { pointsUsed: this.system.karma.useAlways ? 1 : 0, available: this.system.karma.value, step: this.system.karma.step },
+      devotion: { pointsUsed: ability.system.devotionRequired ? 1: 0, pointsRequired: ability.system.devotionRequired, available: this.system.devotion.value, step: this.system.devotion.step },
+      chatFlavor: ability.name + " Test",
+    } );
+    const roll = await RollPrompt.waitPrompt( edRollOptions, options );
+    this.#processRoll( roll );
+  }
+
+  async getTarget( ability ) {
```
ChristopherSD, line 143: I don't think that this should be part of the Actor Document. Since all  the logic depends on the system data of the ability, it should go there.

Generally as a rule of thumb: If you don't need to use `this` in the function, it probably doesn't need to be a class method. (Except static methods which have some other connection to the class)

module/documents/actor.mjs
```diff
@@ -118,6 +118,86 @@ export default class ActorEd extends Actor {
...
+        
+      }
+
+      /**
+       * @param { Array } targets array of all targets
+       * @param { string } targetDefenseType defense 
+       * @param { any } aggregate ???
+       * @returns { number} return  
+       */
+      function _getAggregatedDefense( targets, targetDefenseType, aggregate = Math.max ) {
```
ChristopherSD, line 194: Doesn't belong in the Actor Document. This is some static method for combat or interaction workflows. We need to think where it makes sense.

Generally as a rule of thumb: If you don't need to use `this` in the function, it probably doesn't need to be a class method. (Except static methods which have some other connection to the class)

ChristopherSD (commented):
Sorry for the split reviews. This one concerns the main functionality.

module/data/other/roll-options.mjs
patrickmohrmann: the user can reduce the devotion points in the prompt

module/documents/actor.mjs
patrickmohrmann: Oh i think you did taht acutally, I just added a space for lint sake :-P

module/data/item/templates/none-namegiver-power.mjs
patrickmohrmann: since powerStep and DamageStep are not in ability

module/data/item/talent.mjs
patrickmohrmann: okidoki

templates/item/item-partials/item-details/details/item-details-talent.hbs
patrickmohrmann: cleared

templates/item/item-partials/item-details/details/item-details-attack.hbs
patrickmohrmann: will be refactored

ChristopherSD (approved):

Issue #429, opened 2024-01-19
Workflow Combat - Roll Attack


patrickmohrmann:
combat is triggered by the following options:

1. Attack button 
2. ability use

the attack button will use the weapon equipped in the main hand and roll a dexterity based attack

the ability use will be triggered by the ability itself. Abilities shall become an identifier marking them as combat talents (melee, unarmed, thrown or projectile) 
in addition abilities shall have an option to declare the used equipped hands (main hand, off hand, tail, none) 

when an ability with an combat identifier is rolled the related equipped weapon is used

e.g. 

melee weapons => setting (combat ability = melee, equipped hand = main hand) 
second weapon => setting (combat ability = melee, equipped hand = off hand) 
second attack => setting (combat ability = melee, equipped hand = main hand) 
unarmed attack => setting (combat ability = unarmed, equipped hand = none) 
projectile weapon => setting (combat ability = projectile, equipped hand = main) 
Charge => setting (combat ability = melee, equipped hand = main hand) 
etc.

the attack will be against the selected defense rating of that ability.

Every action will get its own rollData. this will be RollAttackData

Chat options. 
on successful roll search the defender for reaction abilities and offer them in the chat as options. each option gets a button. those multiple options could be folded/unfolded (optional I guess there are not so many options anyway). 
one option is to "not defend".
clicking on one of these buttons will trigger an action for the defending actor.

Another area with a list of potential maneuvers will be available if extra successes are available. In addition a roll damage button shall be available and another button for Assign effect

all options from Reactions, roll damage, apply effect shall be made available for multi target selection

### Optional
when an ability is used and the equipped weapons do not match up, but there is at least one weapon available in the items the actor ownes. a pop up shall appear, requesting the character to re-equip and progress with the newly equipped weapon from here on.

patrickmohrmann:
done

PR #636, opened 2024-03-22
updated weaponType on weapon sheet

### LINKED ISSUE #629: creaet Weapon item sheet and data model
**Description**
create data model and sheet

add melee/thrown to the weapon type option.

**Additional context**

patrickmohrmann:
closes #629

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -421,6 +421,13 @@
                     "strengthMinimum":                          "Min Str",
                     "weaponType":                               "Kategorie"
                 }
+            },
+            "WeaponType": {
+                "melee":				                        "Nahkampfwaffe",
+	            "unarmed":			                            "Waffenlos",
+	            "missle":				                        "Projektilwaffe",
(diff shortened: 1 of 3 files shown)


ChristopherSD (commented):
What is the "meleeThrown" weapon type for?

And just wondering if "unarmed" is needed in the new combat workflow? Are there any other use cases than talent as damage (claw shape, body control)?
In theory the talent could have a a rollType field where you can indicate that it's a damage talent?

patrickmohrmann:
meleeThrown is for weapons like daggers and knifes, which *can* be thrown or used in melee.

the weapon type is used for a lot more than just damage, if we go along with #429 than this will trigger an potential re-equip, checks if equipment is available, used for Damage rolls or, and this is probably most important, defines if a reaction is possible or not (e.g. Reposte is only possible against weapons as is parry etc.) so there are a bunch of use cases why i would like to keep unarmed,

PR #702, opened 2024-05-03
non rollable abilities can still use strain

patrickmohrmann:
Non usabale abilities, don't have an dice symbol anymore but a blood drop to take strain damage when clicking.


ChristopherSD (approved):
Reviewed together in weekly

PR #733, opened 2024-05-29
token update

patrickmohrmann:
ich würde gerne noch folgendes tun:

image

 if ( number === 0 ) color = Color.fromRGB([(1-(pct/2)), pct, 0]);
    else color = Color.fromRGB([(0.5 * pct), (0.7 * pct), 0.5 + (pct / 2)]);

**zusätzlich**
die berechnung scheint nicht an der richtigen stelle zu sein. wenn ich refreshe und den character auf die scene ziehe hat er keine balken, erst nachdem ich den sheet geöffnet habe


module/data/actor/templates/sentient.mjs
```diff
@@ -37,6 +37,22 @@ export default class SentientTemplate extends CommonTemplate {
                 initialKeysOnly: true,
                 label: "ED.Attributes.attributes"
             } ),
+            healthRate: new fields.SchemaField( {
```
ChristopherSD, line 40: What is that for? The token bar values? If so, this should probably go into a `tokenData` `SchemaField` or something like that. So it's clear what it stands for.
Also, why is that necessary for tokens? We have the health ratings elsewhere don't we?

module/hooks/init.mjs
```diff
@@ -22,6 +22,7 @@ export default function () {
         // record configuration values
         CONFIG.ED4E = ED4E;
         CONFIG.Actor.documentClass = documents.ActorEd;
+        CONFIG.Actor.documentClass = documents.CharacterEd;
```
ChristopherSD, line 25: This feels overkill for that one precreate. Also, this Overrides the line before, which we do not want.

module/documents/character.mjs
```diff
@@ -0,0 +1,28 @@
+import ActorEd from "./actor.mjs";
+
+export default class CharacterEd extends ActorEd {
+
+    async _preCreate( options, userId ) {
```
ChristopherSD, line 5: I don't think we need an extra subclass for that one thing, since we have to have the typeguard anyways. We can use Polymorphisms, but then we should do it systematically, like here: https://foundryvtt.wiki/en/development/guides/polymorphism-actors-items

ChristopherSD (dismissed):

patrickmohrmann:
token color. im init tokenDocument Class durch TokenED erweitern und dann wie einen normalen document extenden.

patrickmohrmann:
@ChristopherSD adapted the branch and I reviewed it.

Issue #752, opened 2024-06-11
Jump up ED-id


patrickmohrmann:
**Description**
create an ED-id to enable Abilities to do a jump-up roll 
- abilities example Knack **Rising Handspring** from deeper secrets
- 

**Additional context**

PR #757, opened 2024-06-14
26 recovery workflow

### LINKED ISSUE #26: recovery Workflow
## Reason description

damage and wounds received will in general be heald by recovery tests (there are only a few exceptions)

## detail description

the following tests exist:
- [ ] standard recovery test 
- [ ] Ability based recovery (ability test like Fire blood)
- [ ] Item based recovery (recovery test with potential bonus like Healing potion)
- [ ] recovery for stun damage

## Technical notes

recovery tests count as **Effect Test** with some specials cases (see blow)

calculations:

- for every 6 points in toughness the character receives 1 additional recovery test. (1-6 = 1; 7-12 = 2 etc)
- several bonuses can increase the amount of recovery tests per day.
- a long rest will reset recovery test
- abilities, potions and other effects can increase the recovery step

**Note!**
not every time a recovery test is **used to roll** something, it is actually counted as recovery test. the following rules apply:

it is a recovery test when:
- roll standard recovery
- use potions (healing, last change etc)

it is not a recovery test when:
-  it states **this action uses one recovery test** (see wood skin or fire blood)

### Proposed Solution

include a recovery parameter at each of the following items:
- equipment
- weapon
- armor
- shield
- talent
- skill
- devotion
- knack
- spell

the recovery parameter shall have the following values:

| recovery property Value  | description | Recovery required | Step | roll type|
| ------------- | ------------- | ------------- | ------------- | ------------- |
| 0  | not a healing item | no  | no  | none  |
| 1  | pure healing  | no | arbitrary step  | recovery |
| 2  | Booster Potion  | yes  | arbitrary step + Toughness  | recovery  |
| 3  | Healing Potion | yes/no | arbitrary step + Toughness or arbitrary step  | recovery   |
| 4  | Fire blood  | yes | ability test  | recovery ability |
| 5 | Wood skin  | yes  | ability test  | ability  |

### roll type explaination 
recovery: **Effect test**, wounds only reduce the result not the step. Result reduces damage
healing ability: **Action test**, wounds reduce the step. Result reduces damage
ability: **Action Test** no healing. 

recovery test itself shall be counted as recovery roll type as well

### recovering from stun damage

if a character has stun damage and is rolling a test of type **recovery** or **recovery ability**, an option shall appear in the option box to add the **willpower step** to the **toughness step**. If this is checked, the recovery test will only heal subdual damage (calculation as usual).

**Note!** this option shall only be available if the actor took stun damage since his last recovery test.

patrickmohrmann:
fixes #26 & #752 

this introduces several new options:
1: button support for the bottom buttons on the actor sheet
- Recovery
- knockdown
- jump up
- take damage

added new ed-id for knock-down and jump-up to cover abilities which will be attribute substitutes for knockdown or jump up tests.

## Take Damage

take damage button as several options:
damage value
ignore armor (boolean)
armor type (radio) physical / mystical
damage type (radio) stun / Standard

rules:
standard damage:
- armor counts 
- against standard health
- wound if damage >= than wound threshold --> then knockdown test trigger

stun damage: 
- armor counts
- against stun damage
- harried = true if damage >= wound threshold --> then knockdown test trigger

ignore armor does exactly that

armor type does exactly that.

## Knock down test

checks for abilities which have the knock-down edid. (creates a dynamic pop up --> styling missing)
if found they will be substitude strength.
test against taken damage - woundthreshold
fail => causes knockdown.

## jump up test

checks for abilities which have the jump-up edid. (creates a dynamic pop up)
causes 2 strain against DC 6
if ability is used, strain will be taken from ability
on success => knocked down flag removed

## recovery test

three types of recovery 
1. standard
2. stun
3. full nights rest

stun recovery can only be used after stun damage was taken
stun recovery is rolled with will power step + toughness step
regenerates only stun damage
only one stun recovery test can be taken (this is solved by a flag which will be set to true after the stun recovery test is rolled) all other recovery tests will reset this flag
uses one recovery test
result is reduced by the number of wounds

standard recovery regenerates both for the same amount (this is an open question i have with fasa)
uses one recovery test
result is reduced by the number of wounds

full nights rest will regenerates both for the same amount (this is an open question i have with fasa)
resets recovery tests if no damage/wounds
uses one recovery test
result is reduced by the number of wounds
if damage is 0 it will instead heal 1 wound


module/applications/actor/base-sheet.mjs
```diff
@@ -73,12 +73,28 @@ export default class ActorSheetEd extends ActorSheet {
     // Owned Item management
     html.find( ".card__ability .take-strain" ).click( this._takeStrain.bind( this ) );
 
-    // Owned Item management
-    html.find( ".card__ability .take-strain" ).click( this._takeStrain.bind( this ) );
-
     // Owned Item management
     html.find( ".item-delete" ).click( this._onItemDelete.bind( this ) );
 
+    // Actor Sheet Buttons
```
ChristopherSD, line 79: I would suggest to add a data attribute to these elements on the basis of which the appropriate function is chosen. So you only need to find one class, like ".rollable". So, e.g., `<span class=".rollable" data-action="recovery">`. And the function is something like:
```javascript
html.find( ".rollable" ).click( event => {
  const action = event.currentTarget.dataset.action;
  actionMapping = {
    "recovery": this._onRecoveryRoll( event ),
    "takeDamage": this._onTakeDamage( event ),
    // or, if only one line:
    "jumpUp": this.actor.jumpUp(),
    // etc....
  };
} );
```

module/data/actor/templates/sentient.mjs
```diff
@@ -135,7 +135,7 @@ export default class SentientTemplate extends CommonTemplate {
                             required: true,
                             nullable: false,
                             min: 0,
-                            step: 1,
+                            // step: 1,
```
ChristopherSD, line 138: Any reason this is commented out?

module/data/actor/templates/sentient.mjs
```diff
@@ -153,9 +153,9 @@ export default class SentientTemplate extends CommonTemplate {
...
```
ChristopherSD, line 156: Any reason this is commented out?

module/data/actor/templates/sentient.mjs
```diff
@@ -153,9 +153,9 @@ export default class SentientTemplate extends CommonTemplate {
...
                             initial: 0,
-                            integer: true,
+                            // integer: true,
```
ChristopherSD, line 158: Any reason this is commented out?

module/data/actor/templates/sentient.mjs
```diff
@@ -187,11 +187,16 @@ export default class SentientTemplate extends CommonTemplate {
                         required: true,
                         nullable: false,
                         min: 0,
-                        step: 1,
+                        // step: 1,
```
ChristopherSD, line 190: Any reason this is commented out?

module/data/actor/templates/sentient.mjs
```diff
@@ -187,11 +187,16 @@ export default class SentientTemplate extends CommonTemplate {
...
                         initial: 0,
                         integer: true,
                         label: "ED.Actor.Characteristics.recoveryTestsCurrent"
                     } ),
+                    stun: new fields.BooleanField( {
```
ChristopherSD, line 195: Does that mean, that you're attacking for stunning? If so, maybe take a more descriptive name.
Also, shouldn't that be part of condition, like e.g. aggressive attack?

module/dice/ed-roll.mjs
```diff
@@ -71,6 +71,7 @@ export default class EdRoll extends Roll {
       ? Math.max( super.total, 1 )
       : super.total;
   }
+
```
ChristopherSD, line 74: :D

module/documents/actor.mjs
```diff
@@ -76,27 +81,50 @@ export default class ActorEd extends Actor {
...
    * Legend point History earned prompt trigger
+   * @returns {Promise<void>}
    */
-  async legendPointHistoryEarned( ) {
+  async legendPointHistoryEarned() {
     // let history = await getLegendPointHistoryData( actor );
-    const lpUpdateData = await LegendPointHistoryEarnedPrompt.waitPrompt( new LpTrackingData( this.system.lp.toObject() ), {actor: this} );
-    return this.update( {system: { lp: lpUpdateData }} )
+    const lpUpdateData = await LegendPointHistoryEarnedPrompt.waitPrompt( new LpTrackingData( this.system.lp.toObject() ), { actor: this } );
+    return this.update( { system: { lp: lpUpdateData } } )
```
ChristopherSD, line 89: The doc says it returns `Promise<void>` which is not true I think. It's the updated document iirc.

module/documents/actor.mjs
```diff
@@ -76,27 +81,50 @@ export default class ActorEd extends Actor {
...
   }
 
+  // eslint-disable-next-line max-params
+  async createEdRollOptions( testType = "action", rollType, strain = {}, target = {}, step = {}, devotionRequired, chatFlavor ) {
```
ChristopherSD, line 93: Why is this async?

Also, this should probably be a `static` method of de `EdRollOptions` where you pass in this document...

Also, the signature should be shortened, where you're just passing the basic data in an object. Something like:
```javascript
/**
 @param options {edRollCreationOptionsblabla} object with options like `devotionRequired` or sth like that
 */
static createEdRollOptions( data, actor, options ) {
  // do stuff
}
```

module/documents/actor.mjs
```diff
@@ -76,27 +81,50 @@ export default class ActorEd extends Actor {
...
+    let devotion = {}
```
ChristopherSD, line 94: Forgot closing semicolons here and in rest of function.

module/documents/actor.mjs
```diff
@@ -76,27 +81,50 @@ export default class ActorEd extends Actor {
...
+    if ( devotionRequired === true ) {
```
ChristopherSD, line 95: Can be shortened (including line above):
```javascript
devotion = { 
  pointsUsed: devotionRequired ? 1: 0, 
  available: this.system.devotion.value,
  step: this.system.devotion.step,
}
```

module/documents/actor.mjs
```diff
@@ -115,84 +143,267 @@ export default class ActorEd extends Actor {
...
+      sourceActor: this.name,
+      equipment: equipment.name,
+      step: arbitraryStep
     } );
+    const edRollOptions = await this.createEdRollOptions( "action", "equipment", 0, difficultyFinal, arbitraryStep, false, chatFlavor );
     const roll = await RollPrompt.waitPrompt( edRollOptions, options );
     this.#processRoll( roll );
   }
 
+  async takeDamageManual( options = {} ) {
```
ChristopherSD, line 183: Why is this necessary? For the prompt?
That can be incorporated in the `takeDamage` function. Or rather: make it like the other prompts: call it from the listener and pass the data to the `takeDamage` function

module/documents/actor.mjs
```diff
@@ -201,25 +412,131 @@ export default class ActorEd extends Actor {
    * <ul>
    *     <li>taking strain damage</li>
    *     <li>reducing resources (karma, devotion)</li>
+   *     <li>recover from damage</li>
    * </ul>
    * @param {EdRoll} roll The prepared Roll.
+   * @param {string} recoveryMode type of recovery
    */
-  #processRoll( roll ) {
+  #processRoll( roll, recoveryMode ) {
```
ChristopherSD, line 420: This is way to specific a parameter for a function that should be generalized.
We need to find a way of processing this another way

module/documents/actor.mjs
```diff
@@ -265,22 +582,22 @@ export default class ActorEd extends Actor {
...
+      } )
       );
     }
   }
 
   async addLpTransaction( type, transactionData ) {
     const oldTransactions = this.system.lp[type];
-    const transactionModel = type === "earnings" ? LpEarningTransactionData : LpSpendingTransactionData
-    const transaction = new transactionModel( transactionData )
+    const TransactionModel = type === "earnings" ? LpEarningTransactionData : LpSpendingTransactionData
```
ChristopherSD, line 597: Closing semicolons messing in multiple lines

module/config.mjs
```diff
@@ -113,6 +113,12 @@ ED4E.globalBonuses = {
 	allRangedDamage: {
 		label: 			"ED.Actor.GlobalBonus.allRangedDamage"
 	},
+	allRecoveryEffects: {
```
ChristopherSD, line 116: Can we try to introduce some code internal distinction between recovery test the process/test vs. recovery test the ressource?

With this name this would in my mind then include things that use the ressource?

module/config.mjs
```diff
@@ -113,6 +113,12 @@ ED4E.globalBonuses = {
...
+		label: 			"ED.Actor.GlobalBonus.allRecoveryEffects"
+	},
+	allKnockdownEffects: {
```
ChristopherSD, line 119: Are there multiple knockdown effects?

module/utils.mjs
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-params */
```
ChristopherSD, line 1: Which functions causes this warning? Can we improve it?

ChristopherSD (changes requested):
This is a first scan of the changes.
I did not yet review the actual big functions that do the things!

Also, I would like to start working with AppV2 and the automated input generation from data models as soon as possible. These small prompts are perfect for this.

module/applications/actor/base-sheet.mjs
patrickmohrmann: Clever

module/data/actor/templates/sentient.mjs
patrickmohrmann: yes, correct put it back in place

module/data/actor/templates/sentient.mjs
patrickmohrmann: yes, correct put it back in place

module/data/actor/templates/sentient.mjs
patrickmohrmann: yes, correct put it back in place

module/data/actor/templates/sentient.mjs
patrickmohrmann: yes, correct put it back in place

module/data/actor/templates/sentient.mjs
patrickmohrmann: no, it does not, there is a special rule for recovering from stun damage, but you can do that only once after receiving stun damage. I had the idea to change this boolean after each recovery check. One hour has to be between each recovery checks, so I thought that would be okay.

stun recovery is adding will powe step to toughness step but regenerates only stun damage

module/dice/ed-roll.mjs
patrickmohrmann: thats me trying to place some easter eggs

module/documents/actor.mjs
patrickmohrmann: ok, unsure what this should be

module/documents/actor.mjs
patrickmohrmann: I don't really understand what this would mean

module/documents/actor.mjs
patrickmohrmann: how do I avoid calling the prompt if i incorporate it into takeDamage? that was my reason to exclude it

module/documents/actor.mjs
patrickmohrmann: it would be a sub set of a subset:

testType = effect
rollType = recovery
recoveryMode = stun, standard, nightrest

check together

module/documents/actor.mjs
patrickmohrmann: can we make it an eslint ?

module/config.mjs
patrickmohrmann: unsure, this effect will add x step to every recovery you do. we can certainly add "resource" to all Resources we have

module/config.mjs
patrickmohrmann: you can have threads providing bonus to tests agains knockdown from group pattern, items and what not

module/utils.mjs
patrickmohrmann: lets check together

module/data/actor/templates/sentient.mjs
patrickmohrmann: Bessere Bezeichnung

module/documents/actor.mjs
patrickmohrmann: Promise<void> returns null or undefined...

module/documents/actor.mjs
patrickmohrmann: image

module/documents/actor.mjs
patrickmohrmann: takeDamageManual in den BaseSheet am listener und nach prompt dann ganz normal takeDamage im actor aufrufen.

module/documents/actor.mjs
patrickmohrmann: subset ist richtig

module/config.mjs
patrickmohrmann: Rename resourcen into recoveryResource.

module/config.mjs
patrickmohrmann: ja es ist **kein globaler Bonus** es ist ein **individueller Bonus** es ist kein Fähnchen sondern in der Auswahl der Boni

patrickmohrmann: und bitte auch einbauen Herr Mohrmann nich tnur anllegen

module/utils.mjs
patrickmohrmann: ja ist nicht für alle sondern nur einzelne,

module/data/actor/templates/sentient.mjs
patrickmohrmann: renamed in stunRecoveryAvailability

module/documents/actor.mjs
patrickmohrmann: added semi: ["error", "always"] under rules

PR #811, opened 2024-07-13
Refactoring and smaller fixes for Recovery process branch

ChristopherSD:
Migrate to use AppV2/DialogV2 for the created prompts. 

Fix smaller bugs, like knockdown trigger on taking damage (must be threshold + 5).

Simplify the code a bit.

Lint a lot. Also, prettier is now removed, it just makes things unnecessarily complicated...


module/documents/actor.mjs
```diff
@@ -374,71 +362,65 @@ export default class ActorEd extends Actor {
     let strain = 0;
     let knockdownStep = attributes.str.step;
 
-    const knockdownAbilities = this.items.filter( item => item.system.edid === "knock-down" );
-
-    if ( knockdownAbilities.length > 0 ) {
-      const knockdownAbilityId = await KnockDownItemsPrompt.waitPrompt( knockdownAbilities );
+    const knockdownAbility = await fromUuid(
```
patrickmohrmann, line 365: @ChristopherSD this returns always null even though i have three items with knock-down ed id. The buttons are there, but the abilities are not used.

patrickmohrmann: wierd thing... the step is one higher than the attribute step even though I don't have an ability

module/documents/actor.mjs
```diff
@@ -624,14 +604,13 @@ export default class ActorEd extends Actor {
       }
 
       return changes.concat(
-
         e.changes.map( ( c ) => {
           // eslint-disable-next-line no-param-reassign
-          c = foundry.utils.duplicate( c );
+          c = futils.duplicate( c );
```
patrickmohrmann, line 609: @ChristopherSD  what doest _applyBaseEffects do?

module/documents/actor.mjs
```diff
@@ -374,71 +362,65 @@ export default class ActorEd extends Actor {
...
       return;
     }
     const { attributes } = this.system;
     let devotionRequired = false;
     let strain = 2;
     let jumpUpStep = attributes.dex.step;
 
-    const jumpUpAbilities = this.items.filter( item => item.system.edid === "jump-up" );
-
-    if ( jumpUpAbilities.length > 0 ) {
```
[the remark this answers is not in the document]
ChristopherSD: I don't get it. If your ability has an attribute assigned it should not take the jump up step level, that's correct. I assume your abilities all have an attribute value?
This is just a failsafe in case there is an ability that doesn't rely on an attribute and just add its level flat on the test...

module/documents/actor.mjs
ChristopherSD: No idea. I thought you copied in here from the last system? XD 
This function was already there...

module/documents/actor.mjs
```diff
@@ -448,19 +430,18 @@ export default class ActorEd extends Actor {
...
+        target: difficulty,
+        step: jumpUpStepFinal,
+        devotionRequired: devotionRequired,
+        chatFlavor: chatFlavor
       },
-      this,
-     );
+      this
+    );
     const roll = await RollPrompt.waitPrompt( edRollOptions, options );
```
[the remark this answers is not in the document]
ChristopherSD: I'm not sure what you mean. I didn't change anything.
The way you did it is good!

module/documents/actor.mjs
patrickmohrmann: :-P I'll have a look

PR #812, opened 2024-07-14
added documentation for token settings

patrickmohrmann:


module/documents/actor.mjs
```diff
@@ -29,6 +29,13 @@ export default class ActorEd extends Actor {
     return this.items.filter( item => item.type === 'namegiver' )[0];
   }
 
+  /**
```
ChristopherSD, line 32: I'd strongly recommend to use @inheritdoc here, since we don't change the functionality of the superclass function, and we'd have to adapt it if the super function changes..
I think you can still add the userfunction tag additionally.

module/canvas/token.mjs
```diff
@@ -13,6 +13,14 @@ export default class TokenEd extends Token {
     return super._drawBar( number, bar, data );
   }
 
+  /**
+   * @description             Draw a health bar for the Token representing a rate of health points.
```
ChristopherSD, line 17: damage points

module/canvas/token.mjs
```diff
@@ -13,6 +13,14 @@ export default class TokenEd extends Token {
...
+   * @param {Number} number   Index of the Bar being drawn
+   * @param {Object} bar      PIXI.Graphics instance for the Bar
+   * @param {Object} data     Data object for the Bar
+   * @returns 
+   * @UerFunction             UF_TokenResouces-drawHealthBar
```
ChristopherSD, line 22: missing s: @UserFunction

module/canvas/token.mjs
```diff
@@ -43,6 +51,14 @@ export default class TokenEd extends Token {
     return undefined;
   }
 
+  /**
+   * @description             Draw a karma bar for the Token representing a rate of karma points.
+   * @param {Number} number   Number of the bar being drawn
```
ChristopherSD, line 56: In the upper doc it was "index" of the bar... Should be consistent if possible :)

ChristopherSD (changes requested):

module/documents/actor.mjs
patrickmohrmann: just like this?

 /**
   * @inheritDoc  
   * @description        Pre-process a creation operation of the actor document to ensure that a prototype token is
   * @UserFunction      UF_TokenResouces-preCreate
   */

ChristopherSD (approved):

PR #866, opened 2024-07-31
859 weapons item sheet and card upgrade

patrickmohrmann:
a lot of documentation is part of this issue which is related to Physical items. The documentation has no direct link to the smaller changes i made, I just forgot to switch the branch and don't know /want to (because i fear everything will break :-) ) try to split it.


module/data/item/templates/physical-item.mjs
```diff
@@ -109,6 +109,10 @@ export default class PhysicalItemTemplate extends SystemDataModel.mixin(
                       integer: true,
                       label: "ED.Item.General.arbitraryStep"
                   } ),
+                  action: new fields.StringField( {
```
ChristopherSD, line 112: This should use `ED4E.action` as `choices`

module/data/item/equipment.mjs
```diff
@@ -19,12 +19,14 @@ export default class EquipmentData extends PhysicalItemTemplate.mixin(
                 label: "ED.Item.Equipment.consumable"
             } ),
             // different ammo types are availabel see issue #
-            ammoType: new fields.StringField( {
+            ammunition: new fields.SchemaField( {
```
ChristopherSD, line 22: I can still find 3 occurrences of "ammoType" on this branch... They should be corrected

module/documents/actor.mjs
```diff
@@ -236,6 +237,13 @@ export default class ActorEd extends Actor {
     this.#processRoll( roll );
   }
 
+  /**
+   * 
+   * @param {Object}    itemId        Id of the item to rotate the status of
```
ChristopherSD, line 242: Should probably have a description of the sequence that is rotated (I didn't document it, oops....)

module/handlebar-helpers.mjs
```diff
@@ -97,4 +96,46 @@ export default function registerHandlebarHelpers() {
...
+  /*            Handlebars for Item sheets             */
+  /*************************************************** */
+  Handlebars.registerHelper("isRangedWeapon", function(weaponType) {
+    // Define the logic to determine if the weapon is ranged
+    const rangedWeaponTypes = ["bow", "crossbow", "blowgun", "thrown"]; // Example types
+    return rangedWeaponTypes.includes(weaponType);
+  })
+
+  /*************************************************** */
+  /*            Handlebars for Item Cards              */
```
ChristopherSD, line 111: These functions are all used only once in a template and way to specific, "single-use". So they probably shouldn't be helpers but just be added in the corresponding `prepareData` methods (this should generally be the first choice, unless it is something very general and often used)

lang/de.json
```diff
@@ -125,6 +125,7 @@
                 "action":                                           "Aktion",
                 "active":                                           "Aktiv",
                 "amount":                                           "Menge",
+                "ammunition":                                       "Munition / Menge",
```
ChristopherSD, line 128: It says `ammunition.type` in the localization calls

ChristopherSD (changes requested):

module/data/item/templates/physical-item.mjs
patrickmohrmann: changed it together with the hbs.

lang/de.json
patrickmohrmann: ok fixed the relevant parts

module/data/item/equipment.mjs
patrickmohrmann: fixed it

module/handlebar-helpers.mjs
patrickmohrmann: removed handlebars handler and included them in the weapons dataModel as getter and the base-sheet for getData

ChristopherSD (approved):

PR #967, opened 2024-09-06
Fix skill collection during Char gen

patrickmohrmann:


ChristopherSD (commented):
in the newest iteration (LP Tracking branch) the tier is a top level property of the ability data model. So that probably needs to be adjusted? Or is this needed anyways when we already do it in the other branch?

patrickmohrmann:
Ill make a fix for the LP branch after that one is through

PR #993, opened 2024-09-18
969 npc datamodel and basic sheet

### LINKED ISSUE #969: NPC DataModel and Basic Sheet
**Description**
set up a DataModel for the Actor Type NPC and add all Data fields to the document sheet.
- add challenging Rate
- add maxWounds 
- update sheet
- remove attribute values

**Additional context**

patrickmohrmann:
resolves #969


module/data/actor/templates/namegiver.mjs
```diff
@@ -25,9 +25,13 @@ export default class NamegiverTemplate extends SentientTemplate {
               // choices need to be in the form of an object with the same key and value
               // for the `formField` Handlebars helper to work correctly and have the name
               // as the value attribute of the option tag
-              choices: () => Object.fromEntries(
-                game.settings.get( "ed4e", "languages" ).map( lang => [ lang, lang ] )
-              ),
+              choices: () => {
+                const languages = game.settings.get( "ed4e", "languages" );
+                if ( !languages || Object.keys( languages ).length === 0 ) {
```
ChristopherSD, line 30: Can be written as 
```javascript
// import section
const futils = foundry.utils;

// code
if ( futils.isEmpty( languages ) ) return {};
```

module/data/actor/templates/sentient.mjs
```diff
@@ -126,6 +126,11 @@ export default class SentientTemplate extends CommonTemplate {
                         integer: true,
                         label: "ED.Actor.Characteristics.woundThreshold"
                     } ),
+                    maxWounds: new fields.NumberField( {
```
ChristopherSD, line 129: Just for me to understand: this means that everything sentient (including PCs, dragons, horrors, spirits, etc) can be a mob? Or should that maybe be a special SchemaField or MixIn that only applies to certain subtypes?

module/data/actor/npc.mjs
```diff
@@ -1,10 +1,15 @@
...
 import NamegiverTemplate from "./templates/namegiver.mjs";
 
+
 /**
  * System data definition for NPCs.
  * @mixin
  */
-export default class NpcData extends NamegiverTemplate {
+export default class NpcData extends NamegiverTemplate.mixin(
+    ActorDescriptionTemplate
```
ChristopherSD, line 11: If this is in here too, then it should be mixed in with the NamegiverTemplate directly, since PC and NPC are the only one extending the namegiver template anyways.

module/data/actor/npc.mjs
```diff
@@ -13,7 +18,38 @@ export default class NpcData extends NamegiverTemplate {
 
     /** @inheritDoc */
     static defineSchema() {
-        return super.defineSchema();
+        const fields = foundry.data.fields;
+        return this.mergeSchema( super.defineSchema(), {
+            mobRules: new fields.BooleanField( {
```
ChristopherSD, line 23: So, only NPCs can be mobs? Then the maxWounds should be here too?

I'd suggest a mob field...
Or at least a more speaking name for this one, like `isMob` (always good practice to name boolean variables with something tha can be asked/answered, imho)

module/data/item/maneuver.mjs
```diff
@@ -1,11 +1,11 @@
-import SystemDataModel from "../abstract.mjs";
 import ItemDescriptionTemplate from "./templates/item-description.mjs";
+import NoneNamegiverPowerData from "./templates/none-namegiver-power.mjs";
 
 /**
  * Data model template with information on Maneuver items.
  * @property {number} extraSuccesses        extra successes to trigger the maneuver
  */
-export default class ManeuverData extends SystemDataModel.mixin(
+export default class ManeuverData extends NoneNamegiverPowerData.mixin(
```
ChristopherSD, line 8: Is that true? NPCs can't have maneuvers? KnackManeuvers dont count here?

module/config.mjs
```diff
@@ -163,6 +163,31 @@ ED4E.action = {
 };
 preLocalize( "action" );
 
+
+/**
+ * challengeRates
+ * @enum {string}
+ */
+ED4E.challengeRates = {
```
ChristopherSD, line 171: I don't think this is necessary... If this is only for localization purposes just make a dynamic localization and call it with `format`, passing the circle number.
Hardcoding numbers like this is too inflexible.

ChristopherSD (changes requested):

module/config.mjs
patrickmohrmann: I don't understand it. lets look at it together

module/data/item/maneuver.mjs
patrickmohrmann: ManeuverData are the maneuver abilities of creatures, like enrage or provoke, they are different to Knack maneuvers. they have the name and the usage of extra successes in common, besides that, the knacks have all the knack stuff on top of it as well. We currently have the differentiation between knackManeuver and Maneuver (may be thats wrong and can be mixxed in later) 
NPCs can have both. maybe the naming of "none-namegiver-powers" is not correct since it is basically for all sentients except characters

module/data/actor/npc.mjs
patrickmohrmann: all none character sentient actors can be mobs, struggled with adding it to sentient since it does not count for all of them

module/data/actor/npc.mjs
patrickmohrmann: i added it to the common template, since all actors have a description

module/data/actor/templates/sentient.mjs
patrickmohrmann: everyone except for characters... i was struggling with this

module/data/item/maneuver.mjs
patrickmohrmann: turned out to be obsolete with update of other branches

module/data/actor/npc.mjs
patrickmohrmann: I added the maxWounds to the sentient model as well, makes sense.

PR #1001, opened 2024-09-24
Living armor2

### LINKED ISSUE #979: Living Armor Only
**Description**
Obsidimen and may be others are limited to wear living armor only. to enable this, Namegiver items shall get an option "living armor only" and Armor items shall get an option "living armor" 

a small function shall throw a error on the switch item status feature if the namegiver setting is true and the armor is not living.

**Additional context**

patrickmohrmann:
resolves #979  and replaces #994

### DIFF
--- a/module/data/item/armor.mjs
+++ b/module/data/item/armor.mjs
@@ -24,66 +24,72 @@ export default class ArmorData extends PhysicalItemTemplate.mixin(
                 armor: new fields.NumberField( {
                 required: true,
                 nullable: false,
-                min: 0,
-                initial: 0,
-                integer: true,
-                label: "ED.Item.Armor.physicalArmor"
+                min:      0,
+                initial:  0,
+                integer:  true,
+                label:    "ED.Item.Armor.physicalArmor"
                 } ), 
                 forgeBonus: new fields.NumberField( {
                     required: true,
                     nullable: false,
-                    min: 0,
-                    initial: 0,
-                    integer: true,
-                    label: "ED.Item.Armor.forgeBonusPhysical"
+                    min:      0,
+                    initial:  0,
+                    integer:  true,
+                    label:    "ED.Item.Armor.forgeBonusPhysical"
                 } ),
             } ),
             mystical: new fields.SchemaField( {
                 armor: new fields.NumberField( {
                     required: true,
                     nullable: false,
-                    min: 0,
-                    initial: 0,
-                    integer: true,
-                    label: "ED.Item.Armor.mysticalArmor"
+                    min:      0,
+                    initial:  0,
+                    integer:  true,
+                    label:    "ED.Item.Armor.mysticalArmor"
                 } ),
                 forgeBonus: new fields.NumberField( {
                     required: true,
                     nullable: false,
-                    min: 0,
-                    initial: 0,
-                    integer: true,
-                    label: "ED.Item.Armor.forgeBonusMystical"
+                    min:      0,
+                    initial:  0,
+                    integer:  true,
+                    label:    "ED.Item.Armor.forgeBonusMystical"
                 } ),
             } ),
             initiativePenalty: new fields.NumberField( {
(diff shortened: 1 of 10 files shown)


module/documents/actor.mjs
```diff
@@ -858,18 +864,23 @@ export default class ActorEd extends Actor {
...
-          const bowAllowed = equippedWeapons[0]?.system.isTwoHandedRanged && itemToUpdate.system.bowUsage;
-          // If there's a two-handed weapon or two one-handed weapons, unequip one
-          const unequipSomeWeapon = equippedWeapons.some( weapon => weapon.system.itemStatus === "twoHands" ) || equippedWeapons.length > 1;
-          if ( !bowAllowed && unequipSomeWeapon ) {
-            // Prefer to unequip off-hand weapon, if available
-            const weaponToUnequip = equippedWeapons.find( weapon => weapon.system.itemStatus === "offHand" ) || equippedWeapons[0];
-            updates.push( { _id: weaponToUnequip.id, "system.itemStatus": "carried" } );
+          // check if namegiver item allows only living armor/shields
+          if ( namegiver.system.livingArmorOnly && itemToUpdate.system.isLivingArmor === false ) {
+            ui.notifications.warn( game.i18n.localize( "ED.Notifications.Warn.livingArmorOnly" ) );
```
ChristopherSD, line 870: See above

module/documents/actor.mjs
```diff
@@ -858,18 +864,23 @@ export default class ActorEd extends Actor {
...
+            break;
+          } else {
```
ChristopherSD, line 872: See above

module/data/item/armor.mjs
```diff
@@ -3,102 +3,108 @@ import ItemDescriptionTemplate from "./templates/item-description.mjs";
...
+      } ),
+      initiativePenalty: new fields.NumberField( {
+        required: true,
+        nullable: false,
+        min:      0,
+        initial:  0,
+        integer:  true,
+        label:    "ED.Item.Armor.initiativePenalty"
+      } ),
+      isLivingArmor: new fields.BooleanField( {
```
ChristopherSD, line 67: Sorry, yet another suggestion 😅 could we just name it `living`? the item already is armor, so we don't need to repeat item types in properties (this should also apply to things like "piecemealArmor" actually..)
It's always difficult to find a good compromise between descriptive property names and well legible code... But I would suggest that in general: properties don't need information about the object their part of in it...

ChristopherSD (changes requested):

module/documents/actor.mjs
```diff
@@ -797,40 +798,45 @@ export default class ActorEd extends Actor {
...
-              // A complete set of piecemeal armor can have up to 5 size points. Armor pieces come in three sizes and
-              // cost a corresponding number of points: large (3), medium (2), and small (1). A set of piecemeal armor
-              // cannot have more than one size of a particular type.
-              const equippedArmor = this.itemTypes.armor.filter( armor => armor.system.itemStatus === "equipped" );
-              const sameSizePiece = equippedArmor.find( armor => armor.system.piecemealArmor.size === itemToUpdate.system.piecemealArmor.size );
-              if ( sameSizePiece ) {
-                updates.push( { _id: sameSizePiece.id, "system.itemStatus": "carried" } );
+          // check if namegiver item allows only living armor/shields
+          if ( namegiver.system.livingArmorOnly && itemToUpdate.system.isLivingArmor === false ) {
+            ui.notifications.warn( game.i18n.localize( "ED.Notifications.Warn.livingArmorOnly" ) );
```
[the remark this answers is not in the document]
patrickmohrmann: good idea!

module/data/item/armor.mjs
patrickmohrmann: perfect!

ChristopherSD (approved):

PR #1028, opened 2024-10-09
Update roll-options.mjs

### LINKED ISSUE #722: Devotion in arbitrary roll from dice symbol
**Describe the bug**
clicking on the dice symbol in the chat, the devotion step is by default set to 4. it should be set to 3 tough.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

patrickmohrmann:
base resource for devotion is not on step 3.

resolve #722

### DIFF
--- a/module/applications/global/roll-prompt.mjs
+++ b/module/applications/global/roll-prompt.mjs
@@ -22,11 +22,6 @@ export default class RollPrompt extends FormApplication {
     this._updateObject( undefined, {
       "step.modifiers.manual": edRollOptions.step.modifiers.manual ?? 0,
     } );
-    if ( edRollOptions.testType === "arbitrary" ) {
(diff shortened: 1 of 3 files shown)


ChristopherSD (commented):
Would love to keep the getter as generic as possible. Suggestion:
remove the `initial` option from the step field in #bonusRessource and instead creat `const karma` and `const devotion` in constructor and assign the initial step there

patrickmohrmann:
added a check in the constructor to decide what the devotion step shall be

ChristopherSD (commented):
Made a new commit with a suggestion that might be more generic and extendable

ChristopherSD (approved):

PR #1057, opened 2024-10-17
Global Notification template

patrickmohrmann:
schaus dir mal an, funktioniert bei jedem roll wenn du karma / devotion einsetzt aber nicht genügend hast

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -826,12 +826,13 @@
       "windling": "Windling"
     },
     "Notifications": {
-      "Error": {},
-      "Info": {
+      "error": {},
+      "info": {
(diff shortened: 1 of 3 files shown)


lang/de.json
```diff
@@ -826,12 +826,13 @@
       "windling": "Windling"
     },
     "Notifications": {
-      "Error": {},
-      "Info": {
+      "error": {},
```
ChristopherSD, line 829: Why are these not capitalized anymore? I thought the rule is that only actual keys (those that have a translation) are written with lower letters?

patrickmohrmann: because, I missed this. probably just added them as small since the ui.notification.warn is small two and i didn't bothered at that moment to make a function to lowercase it

PR #1105, opened 2024-11-02
525 adding versatility talents

### LINKED ISSUE #525: Adding versatility talents
**Description**
If an actor has the versatility talent (ed-id= versatility), it will be checked during dragging a talent to the actor sheet, so that the versatility option for talentcategory is available for the actor.

**Additional context**

patrickmohrmann:
Learn and Increase method requires refactoring @ChristopherSD 

resolves #525 

restructuring the order of the increase steps:
1 Learn or just add
2 choose category
3 choose discipline (if talent)
4 choose tier
5 overview


module/applications/global/prompt-factory.mjs
```diff
@@ -345,6 +357,8 @@ class ActorPromptFactory extends PromptFactory {
 class ItemPromptFactory extends PromptFactory {
 
   _promptTypeMapping = {
+
```
ChristopherSD, line 360: Remove empty line

module/applications/global/prompt-factory.mjs
```diff
@@ -394,6 +408,38 @@ class ItemPromptFactory extends PromptFactory {
...
+      abilityName: this.document.name,
+    } );
+
+    return DialogClass.wait( {
+      rejectClose: false,
+      id:          "choose-tier-prompt",
+      uniqueId:    String( ++globalThis._appId ),
+      classes:     [ "earthdawn4e", "choose-tier-prompt", "flexcol" ],
+      window:      {
+        title:       titleFlavor,
```
ChristopherSD, line 435: Again, why is this extracted as constant when not used anywhere else? It's not wrong, just preference when it comes to readability

module/applications/global/prompt-factory.mjs
```diff
@@ -433,7 +479,20 @@ class ItemPromptFactory extends PromptFactory {
   }
 
   async _talentCategoryPrompt() {
-    const buttons = Object.entries( ED4E.talentCategory ).map(
+
+    const versatilityId = game.settings.get( "ed4e", "edidVersatility" );
```
ChristopherSD, line 483: Small thing: maybe call it `versatilityEdid` to be even more descriptive. Might be confused with the id of an actual item or something...

module/applications/global/prompt-factory.mjs
```diff
@@ -433,7 +479,20 @@ class ItemPromptFactory extends PromptFactory {
...
+    const actor = this.document.actor;
+    let category = ED4E.talentCategory;
+    if ( actor.items.filter( item => item.system.edid === versatilityId ).length === 0 ) {
```
ChristopherSD, line 486: the actor has a method for that, so can be rewritten something like this:
```javascript
this.document.actor.getSingleItemByEdid( versatilityId, "talent" )
```

module/applications/global/prompt-factory.mjs
```diff
@@ -433,7 +479,20 @@ class ItemPromptFactory extends PromptFactory {
...
+      if ( typeof category === "object" && category !== null ) {
```
ChristopherSD, line 487: Why is all that in here? Are you trying to remove versatility from the categories? What for?

module/data/item/templates/ability.mjs
```diff
@@ -186,6 +186,26 @@ export default class AbilityTemplate extends ActionTemplate.mixin(
...
+    const promptFactory = PromptFactory.fromDocument( this.parent );
+    const tier = await promptFactory.getPrompt( "chooseTier" );
+
+    if ( !tier || tier === "cancel" || tier === "close" ) return;
+
+    const updatedItem = await this.parent.update( {
+      "system.tier": tier,
+    } );
+
+    if ( foundry.utils.isEmpty( updatedItem ) ) {
```
ChristopherSD, line 199: Also feels like something we should make a function out of, we're starting to use it a lot. Something like `warnIfNoUpdate( warnMessageKey )`

module/data/item/templates/ability.mjs
```diff
@@ -210,6 +230,32 @@ export default class AbilityTemplate extends ActionTemplate.mixin(
     if ( !createData?.system?.level ) itemData.system.level = 0;
     const learnedItem = ( await actor.createEmbeddedDocuments( "Item", [ itemData ] ) )?.[0];
 
+    let category = null;
+    if ( item.type === "talent" ) {
```
ChristopherSD, line 234: A line like this is good indication that inheritance should be used. So this entire block should be part of `TalentData`'s `learn` method instead of here.

ChristopherSD (changes requested):
Why is the entire learn method of `TalentData` empty? The code in `AbilityTemplate` should be done there.

module/applications/global/prompt-factory.mjs
```diff
@@ -107,6 +107,16 @@ export default class PromptFactory {
     };
   }
 
+  static get chooseNoDisciplineButton() {
```
ChristopherSD, line 110: Splitting hairs here but :D I think `noDisciplineButton` should be enough, since it could be used in other contexts other than choosing something. The choosing is not part of the Button itself, it just says "noDiscipline"...

module/applications/global/prompt-factory.mjs
```diff
@@ -307,14 +307,26 @@ class ActorPromptFactory extends PromptFactory {
 
   async _chooseDisciplinePrompt() {
     const buttons = await this._getItemButtons( this.document.disciplines, "type" );
+    buttons.push(
```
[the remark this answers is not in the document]
ChristopherSD: Added comment to the line with the new button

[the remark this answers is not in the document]
ChristopherSD: We don't need to declare a `const` here if its not used anywhere else. It can just be pushed into the `buttons` array directly

module/applications/global/prompt-factory.mjs
ChristopherSD: hmm, okay. I think the entire `if ( !versatility) {...}` block can be reduced to something like this:
```javascript
const versatilityItem = this.document.actor.getSingleItemByEdid( ... );
const { versatility, ...categoriesWithoutVersatility } = ED4E.talentCategory;

const buttons = Object.entries( versatilityItem ? ED4E.talentCategory : categoriesWithoutVersatility  ).map(...)
```

module/data/item/templates/ability.mjs
ChristopherSD: Sure, let's do that

ChristopherSD (changes requested):

module/applications/global/prompt-factory.mjs
patrickmohrmann: i do that for readability and debugbability but i think it can be added to the title itself

module/applications/global/prompt-factory.mjs
patrickmohrmann: I don't get that

patrickmohrmann: ok

ChristopherSD (approved):

PR #1107, opened 2024-11-04
draft - extract action workflows from actor.mjs

patrickmohrmann:
take a look if this is the way to extract the workflows from the actor


patrickmohrmann:
so many changes since then

PR #1118, opened 2024-11-08
Update karma and devotion resource on level up

### LINKED ISSUE #269: [Advancement] - add and increase items automatic on level up!
**Description**
if a **level up** is done, the circle / rank of an advancement item increases. If this is done, all items available shall be added to the actor:

### for Disciplines:
- [x] Discipline talent
- [x] Special abilities
- [ ] karma step increase

### for Questors
- [ ] Devotion step increase

### discipline and Path popups
a pop up shall appear providing the option to choose one optional talent from the list of the current and all lower tiers. This list shall only show the once not available to the actor yet

### Questor popup
a pop up shall appear providing the option to choose one devotion from the list of the current and all lower tiers. This list shall only show the once not available to the actor yet.
after this another popup might appear if a devotion ability is set up for this level providing the option to chose from the devotion ability list. This list shall only show the once not available to the actor yet.

if a pop up was closed, the user can open the item and head over to the advancement tab. by tabbing into the empty optional field, the popup shall appear again.

## Attention:

all Active Effects shall be of type permanent 

all choices shall be visible in the advancement tab of that item. 

A check is done prior to the level up, if the Actor has enough free LP available to spend for the discipline talent. Paths are increased based on a knack rank. Same as for questors.

the pop ups will again do a precheck before appearing and will only appear if enough LP are available.

**Additional context**

patrickmohrmann:
resolves #269

### DIFF
--- a/module/data/item/templates/class.mjs
+++ b/module/data/item/templates/class.mjs
@@ -165,11 +165,30 @@ export default class ClassTemplate extends ItemDataModel.mixin(
       nextLevelData.abilities.special.map( ability => fromUuid( ability ) )
     );
 
+
+
     const effects = Array.from( nextLevelData.effects );
 
     await this.parentActor.createEmbeddedDocuments( "Item", [ ...freeAbilityData, ...specialAbilityData ] );
     await this.parentActor.createEmbeddedDocuments( "ActiveEffect", effects );
     // TODO: activate permanent effects immediately
+
+    // increase resource step of the discipline
+
+    const highestClass = this.parentActor.items.filter(
+      item => item.type === "discipline"
+    ).sort(     // sort descending by circle/rank
+      ( a, b ) => a.system.level > b.system.level ? -1 : 1
+    )[0];
+    
+    if ( this.parent.type === "discipline" && this.parent === highestClass ) {
+      const resourceStep = nextLevelData.resourceStep;
+      await this.parentActor.update( { "system.karma.step": resourceStep } );
+    } else if ( this.parent.type === "questor" ) {
+      const resourceStep = nextLevelData.resourceStep;
+      await this.parentActor.update( { "system.devotion.step": resourceStep } );
+    }
+    
 
     // increase all abilities of category "free" to new circle, if lower
     const freeAbilities = this.parentActor.items.filter(
--- a/module/data/item/templates/class.mjs
+++ b/module/data/item/templates/class.mjs
@@ -165,8 +165,6 @@ export default class ClassTemplate extends ItemDataModel.mixin(
       nextLevelData.abilities.special.map( ability => fromUuid( ability ) )
     );
 
-
-
     const effects = Array.from( nextLevelData.effects );
 
     await this.parentActor.createEmbeddedDocuments( "Item", [ ...freeAbilityData, ...specialAbilityData ] );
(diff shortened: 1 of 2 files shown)


module/data/item/templates/class.mjs
```diff
@@ -165,12 +165,31 @@ export default class ClassTemplate extends ItemDataModel.mixin(
       nextLevelData.abilities.special.map( ability => fromUuid( ability ) )
     );
 
+
```
ChristopherSD, line 168: Remove these two empty lines

module/data/item/templates/class.mjs
```diff
@@ -165,12 +165,31 @@ export default class ClassTemplate extends ItemDataModel.mixin(
...
     const effects = Array.from( nextLevelData.effects );
 
     await this.parentActor.createEmbeddedDocuments( "Item", [ ...freeAbilityData, ...specialAbilityData ] );
     await this.parentActor.createEmbeddedDocuments( "ActiveEffect", effects );
     // TODO: activate permanent effects immediately
 
+    // increase resource step of the discipline
+
+    const highestClass = this.parentActor.items.filter(
```
ChristopherSD, line 178: This seems like a good candidate for a dynamic getter (`get highestClass()`) in the document class itself

module/data/item/templates/class.mjs
```diff
@@ -165,12 +165,31 @@ export default class ClassTemplate extends ItemDataModel.mixin(
...
+      item => item.type === "discipline"
+    ).sort(     // sort descending by circle/rank
+      ( a, b ) => a.system.level > b.system.level ? -1 : 1
+    )[0];
+    
+    if ( this.parent.type === "discipline" && this.parent === highestClass ) {
```
ChristopherSD, line 184: Object comparison is not obvious. I'd rather compare their IDs to be sure

ChristopherSD (changes requested):

module/data/item/templates/class.mjs
patrickmohrmann: ok

module/data/item/templates/class.mjs
patrickmohrmann: good idea, did it

module/data/item/templates/class.mjs
patrickmohrmann: good point, did it.

module/documents/actor.mjs
```diff
@@ -64,6 +64,17 @@ export default class ActorEd extends Actor {
     return this.items.filter( item => item.type === "discipline" );
   }
 
+  /**
+   * Returns the highest discipline of an actor
```
ChristopherSD, line 68: That's great! I'd try to slightly improve the doc: I think type is correct, but then you need to adjust the initial value in the reduce to `undefined` instead of `null`. Which is semantically better anyway.
Also, I would be more specific and say that we return the discipline with the highest circle, and if there are multiple with the same, it's the first encountered. No order guaranteed.

module/data/item/templates/class.mjs
```diff
@@ -171,6 +171,17 @@ export default class ClassTemplate extends ItemDataModel.mixin(
     await this.parentActor.createEmbeddedDocuments( "ActiveEffect", effects );
     // TODO: activate permanent effects immediately
 
+    // increase resource step of the discipline
+    const highestClass = this.parentActor.highestDiscipline;
+
```
ChristopherSD, line 176: I would call it `highestDiscipline`, since it's never gonna be a questor class. Also, since it's only used once, you could just call the getter down in the comparison itself.

ChristopherSD (changes requested):

module/documents/actor.mjs
patrickmohrmann: ah right, of course that makes sense

module/data/item/templates/class.mjs
patrickmohrmann: good point

ChristopherSD (approved):

Issue #1081, opened 2024-10-23
Enhance {{toFormGroup}} helper to work with Mapping fields


patrickmohrmann:
**Description**
the helper formInput and formFields shall work with mappingFields (e.g. defenses or attributes)

**Additional context**

ChristopherSD:
This requires implementation of the abstract `DataField` methods [`toInput`](https://foundryvtt.com/api/classes/foundry.data.fields.DataField.html#toInput) and [`toFormGroup`](https://foundryvtt.com/api/classes/foundry.data.fields.DataField.html#toFormGroup) inthe MappingField class.

PR #1126, opened 2024-11-15
Refactor custom data fields

ChristopherSD:
Refactor only, no functional changes. 

For better consistency and modularity, fields are moved to their own module. Each field class is now their own file.


patrickmohrmann:
duplicate of #1081

PR #1135, opened 2024-11-25
Circle information missing in update prompt

### LINKED ISSUE #1104: Circle not shown in upgrade Prompt
**Description**
Image

**Additional context**

patrickmohrmann:
resolved #1104

### DIFF
--- a/module/data/item/templates/class.mjs
+++ b/module/data/item/templates/class.mjs
@@ -3,6 +3,7 @@ import AdvancementData from "../../advancement/base-advancement.mjs";
 import LpIncreaseTemplate from "./lp-increase.mjs";
 import LearnableTemplate from "./learnable.mjs";
 import ClassAdvancementDialog from "../../../applications/advancement/class-advancement.mjs";
+import ED4E from "../../../config.mjs";
 
 /**
  * Data model template with information on "class"-like items: paths, disciplines, and questors.
@@ -76,7 +77,7 @@ export default class ClassTemplate extends ItemDataModel.mixin(
 
   /** @inheritDoc */
   get requiredLpForIncrease() {
-    return 0;
+    return this.calculateRequiredLp();
   }
 
   /* -------------------------------------------- */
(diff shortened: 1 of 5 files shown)


templates/advancement/advancement-requirements.hbs
```diff
@@ -5,6 +5,7 @@
   </fieldset>
 
   {{#each requirementGroups}}
+  {{log "requirementGroups" this}}
```
ChristopherSD, line 8: remove logs

templates/advancement/class-advancement-summary.hbs
```diff
@@ -4,7 +4,8 @@
       {{#if learning}}
         {{localize "ED.Dialogs.Legend.ClassAdvancement.learningClass" type=classItem.type name=classItem.name}}
       {{else}}
-        {{localize "ED.Dialogs.Legend.ClassAdvancement.increaseClassTo" nextLevel}}
+        {{localize "ED.Dialogs.Legend.ClassAdvancement.increaseClassTo"}}{{nextLevel}}
```
ChristopherSD, line 7: should be a `format` if it depends on input. in this case, handlebars, probably something like `{{localize "ED.Dialogs.Legend.ClassAdvancement.increaseClassTo" nextLevel=nextLevel}}`

module/data/item/templates/class.mjs
```diff
@@ -88,6 +89,20 @@ export default class ClassTemplate extends ItemDataModel.mixin(
 
   /* -------------------------------------------- */
 
+  calculateRequiredLp() {
```
ChristopherSD, line 92: I dont think this needs an extra function. This is already what the dynamic getter is for. The getter won't likely get more functionality, so this can all be moved into that...

ChristopherSD (changes requested):

templates/advancement/advancement-requirements.hbs
patrickmohrmann: done

templates/advancement/class-advancement-summary.hbs
patrickmohrmann: your example is not working anything inside of the localize handlebar does not do anything... what is the problem to add the {{nextLevel}}?

module/data/item/templates/class.mjs
patrickmohrmann: true. done

templates/advancement/class-advancement-summary.hbs
patrickmohrmann: ok, got it to work, great!

ChristopherSD (approved):

PR #1200, opened 2024-12-25
solved smaller issues in char gen

### LINKED ISSUE #1085: smaller fixes in Char gen
fixed 
- error message localization
- magic type on spells 
- some window default options for better usage
- added scrollablility to tabs

patrickmohrmann:
resolves #1085

### DIFF
--- a/module/applications/actor/character-generation-prompt.mjs
+++ b/module/applications/actor/character-generation-prompt.mjs
@@ -105,7 +105,6 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
       width:  1000,
       top:    100,
       left:   100,
-      height: 300,
     }
   };
 
--- a/module/applications/actor/character-generation-prompt.mjs
+++ b/module/applications/actor/character-generation-prompt.mjs
@@ -42,33 +42,18 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
       primary: "namegiver-tab",
     };
   }
-
-  // Define a static method to hold the localization paths
-  static getLocalizationPaths() {
-    return {
-      noNamegiver:         "ED.Dialogs.CharGen.Errors.noNamegiver",
-      noClass:             "ED.Dialogs.CharGen.Errors.noClass",
-      attributes:          "ED.Dialogs.CharGen.Errors.attributes",
-      talentRanksLeft:     "ED.Dialogs.CharGen.Errors.talentRanksLeft",
-      skillRanksLeft:      "ED.Dialogs.CharGen.Errors.skillRanksLeft",
-      notFinished:         "ED.Dialogs.CharGen.Errors.notFinished",
-      maxLanguagesToSpeak: "ED.Dialogs.CharGen.Errors.maxLanguagesToSpeak",
-      maxLanguagesToRead:  "ED.Dialogs.CharGen.Errors.maxLanguagesToRead",
-    };
-  }
   
   // #region Error Messages
   static get errorMessages() {
-    const paths = this.getLocalizationPaths();
     return {
-      noNamegiver:         game.i18n.localize( paths.noNamegiver ),
-      noClass:             game.i18n.localize( paths.noClass ),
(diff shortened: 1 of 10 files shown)


module/applications/actor/character-generation-prompt.mjs
```diff
@@ -43,24 +43,46 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
...
   // #region Error Messages
-  static errorMessages = {
-    noNamegiver:      "X.You didn't choose a namegiver. Pretty difficult to be a person then, don't you think?",
-    noClass:          "X.There's no class selected. Don't you wanna be magic?",
-    attributes:       "X. This is just reminder: there are still some unspent attribute points. They will be converted to extra karma.",
-    talentRanksLeft:  "X.There's still some ranks left for your class abilities. Use them, they're free.",
-    skillRanksLeft:   "X.You haven't used all of your skill ranks. Come on, don't be shy.",
-  };
+  static get errorMessages() {
+    const paths = this.getLocalizationPaths();
```
ChristopherSD, line 62: Hmm, I can't really see the advantage of using an additional function here. This is a one time use, and it feels like it makes it actually a bit more difficult to read when trying to find localizations? 
Why not just put the localization keys directly in the old structure?

module/applications/actor/character-generation-prompt.mjs
```diff
@@ -97,39 +120,46 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
...
+      scrollable: [ "" ],
     },
     "equipment-tab": {
-      template: "systems/ed4e/templates/actor/generation/equipment.hbs",
-      id:       "-equipment-tab",
-      classes:  [ "equipment" ],
+      template:   "systems/ed4e/templates/actor/generation/equipment.hbs",
+      id:         "-equipment-tab",
+      classes:    [ "equipment" ],
+      scrollable: [ "" ],
```
ChristopherSD, line 162: Why make the tabs scrollable instead of the window big enough to contain everything? If scrolling can be avoided, I think thats a good thing. Forces you to focus on the most important parts? (Just talking about the tabs themselves, not long lists of stuff like talents or so on...)

module/applications/actor/character-generation-prompt.mjs
```diff
@@ -252,7 +282,7 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
     // Spells
     context.availableSpellPoints = await this.charGenData.getAvailableSpellPoints();
     context.maxSpellPoints = await this.charGenData.getMaxSpellPoints();
-    context.spells = this.spells.filter( spell => spell.system.magicType === this.magicType );
+    context.spells = this.spells.filter( spell => spell.system.spellcastingType === this.magicType );
```
ChristopherSD, line 285: For consistency, I would probably rename the `magicType` field in this class to `spellcastingType` as well?

ChristopherSD (changes requested):

module/applications/actor/character-generation-prompt.mjs
patrickmohrmann: it didn't work to add the localization into the old function, this was how i got it to work

module/applications/actor/character-generation-prompt.mjs
patrickmohrmann: I disagree. the frame has a width of 1000 which is slightly below tablet width and the height is auto, but nevertheless, if the screen is smaller, it will be editable. Good thing about this scrollable feature is, that it only appears, if the dialog is on overflow. Usually there is no scrollbar

module/applications/actor/character-generation-prompt.mjs
patrickmohrmann: I thought about it, but I really like that it is something different. it makes it easier to understand for me. If this is enough for your inner Monk I would like to keep it htat way

ChristopherSD (approved):

PR #1205, opened 2024-12-26
Base for system specific ChatMessage Data and Documents

ChristopherSD:
Create the foundation for ED system specific ChatMessages documents and their DataModels. This currently has no additional functionality but serves as the basis for future enhancements (e.g. Chat Workflows for attacks, etc). The AttackMessage is just an example of how it might work.

### DIFF
--- a/module/data/chat/attack.mjs
+++ b/module/data/chat/attack.mjs
@@ -1,0 +1,31 @@
+import BaseMessageData from "./base-message.mjs";
+
+export default class AttackMessageData extends BaseMessageData {
+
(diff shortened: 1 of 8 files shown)


module/data/chat/attack.mjs
patrickmohrmann, line 1: is the intention to have damage in the same message as the attack? wouldn't that cause the same problem with reactions etc. providing a really big chat message? Just out of curiosity, not saying no to it, though

patrickmohrmann (approved):

PR #1215, opened 2024-12-29
adding talents from the namegiver to the char gen

### LINKED ISSUE #537: Transfer Namegiver Abilities to the Actor
**Description**
on adding a Namegiver item to an Character or NPC Actor,Abilities shall be transferred to the actor as well.

This shall also happen during character Generation

**Additional context**

### LINKED ISSUE #591: add Namegiver abilities to the Character Generation options
**Description**
certain Namegiver have talents as raicial abilities which might be used during character generation for enhancement. Those Talents have to be available in the abilities area (there has to be a third area.

for versatility, it has to appear another optional talents option with as many options as ranks in versatility and these shall be all talents. (only novice??? check char gen)

**Additional context**

patrickmohrmann:
resolves #591 
resolves #537

### DIFF
--- a/module/data/other/character-generation.mjs
+++ b/module/data/other/character-generation.mjs
@@ -109,7 +109,7 @@ export default class CharacterGenerationData extends SparseDataModel {
         ), {
           required:        true,
           initialKeysOnly: true,
-          initialKeys:     [ "optional", "class", "free", "special", "artisan", "knowledge", "general", "language" ],
+          initialKeys:     [ "optional", "class", "free", "special", "artisan", "knowledge", "general", "language", "namegiver" ],
         } ),
       availableRanks: new fields.SchemaField( {
         talent: new fields.NumberField( {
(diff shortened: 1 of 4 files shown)


module/data/other/character-generation.mjs
```diff
@@ -263,6 +263,35 @@ export default class CharacterGenerationData extends SparseDataModel {
...
+  }
+  
+  async getNamegiverAbilities() {
+    const namegiver = await this.namegiverDocument;
+    if ( !namegiver ) return [];
+  
+    const abilitiesUuid = namegiver.system.abilities || [];
+    const abilities = await Promise.all( abilitiesUuid.map( async ( abilityUuid ) => {
+      const ability = await fromUuid( abilityUuid );
+      return ability.type === "talent" ? abilityUuid : null;
```
ChristopherSD, line 289: This only shows talents. But things like "Heat Sight" will be `specialAbility`. So it should probably check if its `system` data model is an instance of `AbilityTemplate`?

module/data/other/character-generation.mjs
```diff
@@ -597,7 +626,7 @@ export default class CharacterGenerationData extends SparseDataModel {
   }
 
   _getAbilityClassType( abilityType ) {
-    const isClass = [ "class", "optional", "free" ].includes( abilityType );
+    const isClass = [ "class", "optional", "free", "namegiver" ].includes( abilityType );
```
ChristopherSD, line 629: This is not a class type... but see my general feedback in the main review.

ChristopherSD (commented):
So... this is just for racial abilities that are also talents and can be raised with the initial rank pool?
That's how it currently is:
image

What about other things like "Heat Sight"? They are not automatically added in the namegiver tab?

module/data/other/character-generation.mjs
patrickmohrmann: no its actually only supposed to show talents, Heat sight etc. and special abilities, have to be listed in the namegiver selection, I'll create an issue for that.

module/data/other/character-generation.mjs
patrickmohrmann: true, but this is all about spending ranks etc. and that is checking for isClass in all the calculations, I thought it reasonable to have it added here

module/data/other/character-generation.mjs
ChristopherSD: If that's done in another issue, okay

ChristopherSD (approved):

PR #1223, opened 2025-01-02
1171 threads data model

### LINKED ISSUE #1171: Threads Data Model
Items of type Thread shall have the following data model:
- **thread Type** (Thread item, Group Thread, Path Thread....)
- **pattern type** = minor, major, or core (only used for physical items though)
- mystic Defense (check if the path should be similar to actors etc... makes probably sense)
- max ranks
- enchantment Pattern link (uuid field) see enchantment version --> not to use yet
- Tier
- analyzed (boolean) --> shows mystic defense, max ranks, Tier and number of level (but not their content by default)
- isKnown (boolean) --> shows tab in item
- Level
  - analyzed (boolean)
  - active (boolean)
  - key information
    - required (boolean) 
    - Question (description field)
    - Answer (description field)
    - known (boolean)
  - Deed 
    - required (boolean) 
    - description (description field)
    - fulfilled (boolean)
  - Effect
    - description (description field)
  - Pool option for Effects

patrickmohrmann:
resolves #1171


module/data/item/templates/threads.mjs
```diff
@@ -0,0 +1,43 @@
...
+ * @property {boolean} isThreadItem is this item a thread item or not
+ * @mixin
+ */
+export default class ThreadTemplate extends ItemDataModel {
+
+  /** @inheritDoc */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    return this.mergeSchema( super.defineSchema(), { 
+      isThreadItem: new fields.BooleanField( {
```
ChristopherSD, line 16: This will be removed right? Since the thread tab will always be presents?

module/data/thread/thread-level.mjs
```diff
@@ -0,0 +1,156 @@
+import { SparseDataModel } from "../abstract.mjs";
+import IdentifierField from "../fields/identifier-field.mjs";
+
+export default class ThreadLevelData extends SparseDataModel {
+  /** @inheritDoc */
+  /** @inheritDoc */
+  static defineSchema() {
+    const fields = foundry.data.fields;
+    return {
+      identifier: new IdentifierField( {
```
ChristopherSD, line 10: This is taken from advancemen level, right? Or is there a specific purpose for the "identifier"?

module/data/item/_module.mjs
```diff
@@ -54,7 +53,6 @@ export {
   SpellKnackData,
   SpellData,
   TalentData,
-  ThreadData,
   WeaponData
 };
```
ChristopherSD, line 58: Add the Thread Template here, like that probably
`export {default as ThreadTemplate} from "./templates/threads.mjs";`

templates/item/item-partials/item-details/other-tabs/threads.hbs
```diff
@@ -0,0 +1,86 @@
...
+  {{#each item.system.threadData.levels}}
+    <fieldset>
+      <div>
+        {{!-- level --}}
+        {{formField this.schema.fields.level name=(concat "system.threadData.levels." @index ".level") value=this.level localize=true}}
+        {{!-- isAnalysed --}}
+        {{formField this.schema.fields.isAnalysed name=(concat "system.threadData.levels." @index ".isAnalysed") value=this.isAnalysed localize=true}}
+        {{!-- isActive --}}
+        {{formField this.schema.fields.isActive name=(concat "system.threadData.levels." @index ".isActive") value=this.isActive localize=true}}
+        {{!-- keyKnowledge isRequired --}}
```
ChristopherSD, line 65: I would surround logical groups here with a `<fieldset>` element as well, and then add a `<legend>` for more structure.
So groups like: "Key Knowledge" "Deed" "Effect" "Base Data" and so on.

module/data/thread/thread-level.mjs
```diff
@@ -0,0 +1,156 @@
...
+          required: true,
+          nullable: false,
+          initial:  false,
+          // label:    this.labelKey( "PhysicalItems.ThreadItem.Deed.isCompleted" ),
+          // hint:     this.hintKey( "PhysicalItems.ThreadItem.Deed.isCompleted" )
+          label:    "ED.Data.Item.Labels.PhysicalItems.ThreadItem.Deed.isCompleted",
+          hint:     "ED.Data.Item.Hints.PhysicalItems.ThreadItem.Deed.isCompleted"
+        } )
+      } ),
+      effect: new fields.StringField( {
```
ChristopherSD, line 114: Maybe add another field like "activeEffect" but for abilites? In case a thread grants new abilities?

ChristopherSD (changes requested):
Label and Hint for mystic defense of the pattern is missing:
image

Suggestion: put the buttons at the very top or bottom and make them not scrollable. So they're always visible and easily accessible. Probably especially for thread weaving very handy.
image

templates/item/item-partials/item-details/other-tabs/threads.hbs
patrickmohrmann: yes, but wanted to do that later, it was already big

module/data/thread/thread-level.mjs
patrickmohrmann: I would like to have the active effect granting the ability, not the item itself, or do you disagree?

module/data/item/templates/threads.mjs
patrickmohrmann: yes, updated it

module/data/thread/thread-level.mjs
patrickmohrmann: ah yes, that was a left over from copy and paste. I removed it

module/data/item/_module.mjs
patrickmohrmann: aha

ChristopherSD (approved):

PR #1238, opened 2025-01-05
added equipments to the char gen

### LINKED ISSUE #1230: use the core pack items in the char gen
use the core items for the equipment section in the char gen

patrickmohrmann:
resolves #1230

### DIFF
--- a/module/applications/actor/character-generation-prompt.mjs
+++ b/module/applications/actor/character-generation-prompt.mjs
@@ -79,7 +79,6 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
       decrease:        this._onChangeAttributeModifier,
       changeSpell:     this._onClickSpell,
       reset:           this._onReset,
-      selectEquipment: this._selectEquipment,
     },
     form:    {
       handler:        CharacterGenerationPrompt.#onFormSubmission,
--- a/module/applications/actor/character-generation-prompt.mjs
+++ b/module/applications/actor/character-generation-prompt.mjs
@@ -21,6 +21,7 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
     this.questors = documentCollections.questors;
     this.skills = documentCollections.skills;
     this.spells = documentCollections.spells;
+    this.equipment = documentCollections.equipment;
 
     this.availableAttributePoints = game.settings.get( "ed4e", "charGenAttributePoints" );
 
@@ -79,6 +80,7 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
       decrease:        this._onChangeAttributeModifier,
       changeSpell:     this._onClickSpell,
       reset:           this._onReset,
+      selectEquipment: this._onSelectEquipment,
     },
     form:    {
       handler:        CharacterGenerationPrompt.#onFormSubmission,
@@ -280,14 +282,7 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
       return acc;
     }, {} );
 
-
-    // Add equipment to the context
-    context.equipment = {
-      armor:     await this.charGenData.getEquipmentItems( "armor" ),
-      equipment: await this.charGenData.getEquipmentItems( "equipment" ),
-      shields:   await this.charGenData.getEquipmentItems( "shield" ),
-      weapons:   await this.charGenData.getEquipmentItems( "weapon" ),
-    };
+    context.equipment = this.equipment;
 
     context.selectedEquipment = this.charGenData.equipment;
 
@@ -435,8 +430,7 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
     }
     if ( foundry.utils.isEmpty( data.languages ) ) delete data.languages;
 
-    // Process selected equipment
-    this.equipment = data.equipment ?? [];
+    
 
     this.charGenData.updateSource( data );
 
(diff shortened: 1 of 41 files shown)


module/applications/actor/character-generation-prompt.mjs
```diff
@@ -572,13 +581,23 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
     result.then( _ => this.render() );
   }
 
+  static _selectEquipment( _, target ) {
```
ChristopherSD, line 584: What is this for? It's not used, is it?

module/utils.mjs
```diff
@@ -182,6 +182,33 @@ export async function getAllDocuments(
 
 }
 
+/**
+ * get all items from the system compendium with the given name
+ * @param {string} compendiumName  The name of the compendium
+ * @param {string} itemNames      The names of the items to retrieve
+ * @returns {Promise<Item[]>}      The items from the compendium
+ */
+export async function getEquipmentItemsFromCompendium( compendiumName, itemNames ) {
```
ChristopherSD, line 191: This is way too specific for a util function. But it can be easily generalized. Probably just call it something like `getCompendiumItemByName` or something like that.
Though I wonder why there would be multiple items by the same name in a well maintained compendium?

ChristopherSD (commented):
I thought we wanted to get the items from the compendium via ID? Or if by name, it should be done via config values. Something like:
```javascript
ED4E.startingEquipment = {
    torch: {
        name: {
            de: "Fackel",
            en: "Torch",
            fr: "Torche",
            // etc
        },
        uuid: <uuid>
    },
};
```

module/applications/actor/character-generation-prompt.mjs
patrickmohrmann: that was a left over which is not needed any more, thanks

module/utils.mjs
patrickmohrmann: ok, renamed it

patrickmohrmann:
changed as you suggested it

module/utils.mjs
ChristopherSD: it's commented out now. Is that intended?

module/data/other/character-generation.mjs
```diff
@@ -6,11 +6,13 @@ import {
   getSingleGlobalItemByEdid,
   mapObject,
   renameKeysWithPrefix,
-  sum
+  sum,
+  // getCompendiumItemByName
```
ChristopherSD, line 10: If not used, remove

module/utils.mjs
ChristopherSD: If not used, can be removed?

module/data/other/character-generation.mjs
```diff
@@ -290,6 +301,61 @@ export default class CharacterGenerationData extends SparseDataModel {
     return abilities.filter( uuid => uuid !== null );
   }
 
+  // async getEquipmentItems( type ) {
```
ChristopherSD, line 304: If not used can be removed?

module/applications/actor/character-generation-prompt.mjs
```diff
@@ -80,6 +79,7 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
       decrease:        this._onChangeAttributeModifier,
       changeSpell:     this._onClickSpell,
       reset:           this._onReset,
+      selectEquipment: this._selectEquipment,
```
ChristopherSD, line 82: Doesn't exist? Not needed?

module/data/other/character-generation.mjs
```diff
@@ -532,6 +598,20 @@ export default class CharacterGenerationData extends SparseDataModel {
     return this.updateSource( { spells: newSpellSet } );
   }
 
+  async addEquipment( equipmentUuid ) {
```
ChristopherSD, line 601: Is this used anywhere?

module/data/other/character-generation.mjs
```diff
@@ -290,6 +301,61 @@ export default class CharacterGenerationData extends SparseDataModel {
...
+  //     weapons:   [ 
+  //       "Dolch",
+  //       "Messer"
+  //     ],
+  //   }[type];
+
+  //   return await getCompendiumItemByName ( compendiumName, itemNames );
+  // }
+
+  async getEquipmentItems( type ) {
```
ChristopherSD, line 342: If this method is called every time the sheet is re-rendered that might be a lot with that `fromUuid` call (just an assumption though). Maybe that should be initialized once, in the constructor or somewhere and saved? Like the other things?

ChristopherSD (changes requested):

module/utils.mjs
patrickmohrmann: ok

module/data/other/character-generation.mjs
patrickmohrmann: ok

module/data/other/character-generation.mjs
patrickmohrmann: ok

module/applications/actor/character-generation-prompt.mjs
patrickmohrmann: ok

module/data/other/character-generation.mjs
patrickmohrmann: nope, its gone

module/data/other/character-generation.mjs
patrickmohrmann: I understand the point, but don't get the how..

ChristopherSD: You can just add a field at the top of the class and assign the object you do in the prompt context. Something like 
```ecmascript 6
equipmentByType = {
  armor:     await this.getEquipmentItems( "armor" ),
  equipment: await this.getEquipmentItems( "equipment" ),
  shields:   await this.getEquipmentItems( "shield" ),
  weapons:   await this.getEquipmentItems( "weapon" ),
};
```

Then you can call it in the context preparation `context.equipment = this.charGenData.equipmentByType`

module/data/other/character-generation.mjs
```diff
@@ -532,6 +559,20 @@ export default class CharacterGenerationData extends SparseDataModel {
     return this.updateSource( { spells: newSpellSet } );
   }
 
+  // async addEquipment( equipmentUuid ) {
```
ChristopherSD, line 562: Still here? I'm not sure, maybe my IDE is just weird?

ChristopherSD (approved):

PR #1244, opened 2025-01-10
show and hide thread data

### LINKED ISSUE #1173: Hide/show level information based on identification
all level of an item of type thread have a field "analyzed" which is boolean and per default set to false. 
the active tab in the thread tab of an item shall blur or hide all information if this field is false.

if the analyzed field is true, the following fields will be visible:

key information (Question)
effect
deed
effect (EAE)

the key information answer will be empty (the content if filled, is only shown to the gm)

if the key information / known field is true, the key information answer will be come visible

patrickmohrmann:
resolves #1173

### DIFF
--- a/module/applications/item/item-sheet.mjs
+++ b/module/applications/item/item-sheet.mjs
@@ -118,6 +118,7 @@ export default class ItemSheetEd extends HandlebarsApplicationMixin( DocumentShe
       system:                 this.document.system,
       options:                this.options,
       systemFields:           this.document.system.schema.fields,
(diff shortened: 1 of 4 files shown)


ChristopherSD (commented):
The `gmOnly` properties of the thread data fields makes it impossible for (trusted) players create their own threads/patterns (e.g. when GMs delegate it to them). We can do that, but I wanna make sure: is that what we want?

patrickmohrmann:
i removed the gmOnly, you are right

ChristopherSD (approved):

PR #1248, opened 2025-01-14
thread weaving action + lp tracking

### LINKED ISSUE #86: Threads to items
thread items use the standard cost table with the item Tier providing a bonus.

By activating a level lp tracking is triggert

patrickmohrmann:
resolves #86


ChristopherSD (commented):
Hm, I think we should have another look at that. I'm not sure having the physical item be lp increasable is the best way to go unfortunately :/

Issue #1141, opened 2024-12-05
NPC Sheet AppV2


patrickmohrmann:
Make the NPC Sheet an AppV2 Application

PR #1251, opened 2025-01-15
NPC actor APPV2

patrickmohrmann:
resovles #1141

### DIFF
--- a/module/applications/global/prompt-factory.mjs
+++ b/module/applications/global/prompt-factory.mjs
@@ -337,7 +337,10 @@ class ActorPromptFactory extends PromptFactory {
 
   async _drawWeaponPrompt() {
     const buttons = await this._getItemButtons( this.document.itemTypes.weapon, "weapon" );
-
+    if ( buttons.length === 0 ) {
+      ui.notifications.info( game.i18n.format( "ED.Notifications.Info.noWeaponAvailable" ) );
+      return;
+    }
     return DialogClass.wait( {
       rejectClose: false,
       id:          "draw-weapon-prompt",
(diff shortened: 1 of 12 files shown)


module/applications/global/prompt-factory.mjs
```diff
@@ -337,7 +337,10 @@ class ActorPromptFactory extends PromptFactory {
 
   async _drawWeaponPrompt() {
     const buttons = await this._getItemButtons( this.document.itemTypes.weapon, "weapon" );
-
+    if ( buttons.length === 0 ) {
+      ui.notifications.info( game.i18n.format( "ED.Notifications.Info.noWeaponAvailable" ) );
```
ChristopherSD, line 341: NPC actor APPV2 ;P

module/data/actor/templates/sentient.mjs
```diff
@@ -228,8 +228,8 @@ export default class SentientTemplate extends CommonTemplate {
       mobRules: new fields.BooleanField( {
         required: true,
         initial:  false,
-        label:    "ED.Data.Actor.Labels.mobRules",
-        hint:     "ED.Data.Actor.Hints.mobRules"
+        label:    "ED.Data.Actor.Labels.isMob",
```
ChristopherSD, line 231: Would be great if the field itself could be called that as well, for consistency. And makes much more sense with it being a boolean

ChristopherSD (commented):
So, I assume the tabs that will not apply (like class or talents for "creature" types) will be hidden later, right?

module/data/actor/templates/sentient.mjs
patrickmohrmann: good point, changed it

ChristopherSD (approved):

PR #1254, opened 2025-01-19
combined powers and attacks

### LINKED ISSUE #1036: combine attacks and powers
**Description**
attacks and powers have the same options. it should be one item, with only a flag to have it better identifiable

**Additional context**

patrickmohrmann:
resolves #1036

there is no real difference between powers and attacks. it supports reducing the amount of items


module/data/item/power.mjs
```diff
@@ -10,7 +10,15 @@ export default class PowerData extends NoneNamegiverPowerData.mixin(
 
   /** @inheritDoc */
   static defineSchema() {
+    const fields = foundry.data.fields;
     return this.mergeSchema( super.defineSchema(), {
+      isAttack: new fields.BooleanField( {
```
ChristopherSD, line 15: I have to admit, I'm not a huge fan of the `isSomeSpecialTypeOfAbility`, sorry. That's what the ed-id is for, imho.
We could introduce one like `creature-attack` or something like that.
And then we can still add a dynamic getter `isAttack` that checks if this has the given ed-id?

patrickmohrmann: I would not want to waste the ed-id for isAttack

ChristopherSD: What do you mean with "waste"? It's unique either way. And what other ed-ids would an attack need? It's already quite specific

patrickmohrmann: ok, got it, build something in

module/data/item/power.mjs
```diff
@@ -24,4 +23,12 @@ export default class PowerData extends NoneNamegiverPowerData.mixin(
     super.migrateData( source );
     // specific migration functions
   }
+
+  /* -------------------------------------------- */
+  /*  Getter                                      */
+  /* -------------------------------------------- */
+  get isCreatureAttack() {
+    const edidCreatureAttack = game.settings.get( "ed4e", "edidCreatureAttack" );
```
ChristopherSD, line 31: Redundant variable, you can just write it in one line :)
Also, there is no check if it actually is an attack, it's just the edid returned.
```ecmascript 6
/**
 *@type {boolean}
  */
get isCreatureAttack() {
  return this.edid === game.settings.get( "ed4e", "edidCreatureAttack" );
}
```

ChristopherSD (changes requested):

module/data/item/power.mjs
patrickmohrmann: great!

patrickmohrmann: One day i will do this automatically :-P

templates/actor/actor-tabs/powers.hbs
```diff
@@ -9,8 +9,10 @@
     <div class="power-card__grid--item text__alignment--center">{{localize "ED.Actor.Header.strain"}}</div>
     <div class="power-card__grid--item text__alignment--center">{{localize "ED.Actor.Header.controlIcons"}}</div>
 </div>
-{{#each actor.itemTypes.attack}}
-    {{> "systems/ed4e/templates/actor/cards/power-card.hbs"}}
+{{#each actor.itemTypes.power}}
+    {{#if (eq this.system.isCreatureAttack this.system.edid)}}
```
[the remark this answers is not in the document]
patrickmohrmann: right

ChristopherSD (approved):

PR #1316, opened 2025-01-20
introduced Group actor and refactored sheets

### LINKED ISSUE #1312: Create Group Actors
Create A Basic DataModel for Actors of Type Group

patrickmohrmann:
I'm starting to refactor the actor applications to have them similar to our data Models. this allows us better to differentiate between options certain actor types do not have and others do.

resolves #1312

### DIFF
--- a/module/applications/actor/namegiver-sheet.mjs
+++ b/module/applications/actor/namegiver-sheet.mjs
@@ -1,0 +1,67 @@
+import ED4E from "../../config.mjs";
+import ActorSheetEdSentient from "./sentient-sheet.mjs";
+
+/**
+ * Extend the basic ActorSheet with modifications
+ * @augments {ActorSheet}
+ */
+export default class ActorSheetEdNamegiver extends ActorSheetEdSentient {
+
+  constructor( options = {} ) {
+    super( options );
+  }
+
+  /** 
+   * @override 
+   */
+  static DEFAULT_OPTIONS = {
+    classes:  [ "earthdawn4e", "sheet", "actor" ],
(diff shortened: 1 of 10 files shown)


module/applications/actor/namegiver-sheet.mjs
```diff
@@ -0,0 +1,67 @@
...
+      positioned:     true,
+      icon:           false,
+      minimizable:    true,
+      resizable:      true,
+    },
+    form: {
+      submitOnChange: true,
+    },
+    actions:  {
+      halfmagic:           ActorSheetEdNamegiver.rollHalfMagic,
```
ChristopherSD, line 30: the `data-action` attribute is called "halfMagic" with a capital "M", so, it doesn't work ;)

ChristopherSD (changes requested):
Currently the "group" type sheet has all the buttons from the other sheets, like "half magic", "recovery", "jump up", etc. Is that intended?

module/applications/actor/namegiver-sheet.mjs
patrickmohrmann: yes it is intended. I want to refactor it further

ChristopherSD: so, it's intended that the half magic button doesn't have functionality after this PR?

patrickmohrmann: yes, it will be removed when i restructure the sheets based on the datamodel

ChristopherSD (approved):

PR #1328, opened 2025-01-21
solved strain issues

### LINKED ISSUE #700: Non-usable Abilities Step 3 - use strain
**Description**
some non-usable abilities might not be used for an actual roll (see #699) but might require Strain. for these particular abilities, a blood drop symbol shall appear while hovering over it (instead of a dice symbol).  

if this blood drop icon is clicked, the strain will be used and a chat message appears with the info, that the ability was used.

--> blood drop is implemented
**chat message ist missing**

**Additional context**

### LINKED ISSUE #756: create chat message for damage taken
**Description**
a chat file for taken damage shall be created and it shall hold an option to do knockdown tests if applicable

**Additional context**

patrickmohrmann:
resolves #756
resolves #700

last two commits are for fixing the linting


module/documents/actor.mjs
```diff
@@ -560,12 +562,25 @@ export default class ActorEd extends Actor {
     }
 
     this.update( updates );
+    let chatFlavor;
+    if ( createChatMessage === undefined ) {
```
ChristopherSD, line 566: This doesn't need to be a whole if else block. The only difference is in the localization. I think it could be sth like this, not tested though:

```ecmascript 6
const chatFlavor = game.i18n.format(
  `ED.Chat.Flavor.${createChatMessage ? "takeStrainDamage", "takeDamage" }`,
  { ability: foo, actor: foo, amount: foo} 
);
```

This is with the assumption that `format` just ignores parameters that don't exist in the localization string.

module/documents/actor.mjs
```diff
@@ -534,13 +535,14 @@ export default class ActorEd extends Actor {
    *                                                            'physical', 'mystical', or 'none'.
    * @param {boolean} [ignoreArmor]                             Whether armor should be ignored when applying this damage.
    * @param {EdRoll|undefined} [damageRoll]                     The roll that caused this damage or undefined if not caused by one.
+   * @param {string} createChatMessage                          The ability name in case of a non rollable ability, or undefined
```
ChristopherSD, line 538: Think about maybe calling it something different. This evokes the feeling (at least for me) that this is a boolean and nothing else. You could just call `abilitName` or `strainOrigin` or whatever and still use it like you do. An empty string or undefined is treated as a falsy value anyways. Just make the parameter optional then

module/documents/actor.mjs
```diff
@@ -534,13 +535,14 @@ export default class ActorEd extends Actor {
...
    * @returns {{damageTaken: number, knockdownTest: boolean}}
    *                                                            An object containing:
    *                                                            - `damageTaken`: the actual amount of damage this actor has taken after armor
    *                                                            - `knockdownTest`: whether a knockdown test should be made.
    */
   // eslint-disable-next-line max-params
-  takeDamage( amount, isStrain, damageType = "standard", armorType, ignoreArmor, damageRoll ) {
+  takeDamage( amount, isStrain, damageType = "standard", armorType, ignoreArmor, damageRoll, createChatMessage ) {
```
ChristopherSD, line 545: we should start thinking about combining the parameters in a configObject to reduce the number of params. Or at least put all the optional params into one and just leave the first three required ones like is.

module/applications/actor/character-sheet.mjs
```diff
@@ -264,6 +264,7 @@ export default class ActorSheetEdCharacter extends ActorSheetEd {
     event.preventDefault();
     const li = target.closest( ".item-id" );
     const ability = this.document.items.get( li.dataset.itemId );
-    this.document.takeStrain( ability.system.strain );
+    const createChatMessage = ability.system.attribute !== "" ? false: ability.name;
```
ChristopherSD, line 267: I think is an unnecessary local variable? Can just be written directly in the function call...

module/documents/actor.mjs
patrickmohrmann: I knew you would say something about it :-P

patrickmohrmann: i like strainOrigin!

module/documents/actor.mjs
patrickmohrmann: i got it to work with my new friend

module/documents/actor.mjs
patrickmohrmann: ok, next time 👍 I create a refactor for it #1331

module/applications/actor/character-sheet.mjs
patrickmohrmann: hä?

module/documents/actor.mjs
ChristopherSD: Yes :) Currrently you're checking if `strainOrigin === undefined`. That works if you're disciplined with how you call the function. If you want it to be more flexible you could just check for a falsy value, e.g. something like `if(!strainOrigin)`, that also applies if someone accidentally sends a boolean or an empty string...
Depends on your intent

module/documents/actor.mjs
```diff
@@ -534,13 +535,14 @@ export default class ActorEd extends Actor {
    *                                                            'physical', 'mystical', or 'none'.
    * @param {boolean} [ignoreArmor]                             Whether armor should be ignored when applying this damage.
    * @param {EdRoll|undefined} [damageRoll]                     The roll that caused this damage or undefined if not caused by one.
+   * @param {string} strainOrigin                               The ability name in case of a non rollable ability, or undefined
```
ChristopherSD, line 538: The name of the parameter needs to be in square brackets so it's clear that it's an optional parameter

module/applications/actor/character-sheet.mjs
ChristopherSD: You can just write:
```ecmascript 6
this.document.takeStrain( 
  ability.system.strain,
  ability.system.attribute 
);
```
If you adapt the function to treat the `strainOriding` as a falsy value not just undefined. Because if the attribute is an empty string, it is automatically `false`.

patrickmohrmann:
last commit was only merge conflicts

module/documents/actor.mjs
patrickmohrmann: I changed it once in the chatFlavor, but in the formula to decide if a message is needed it provides more problems if i check for something else then undefined

module/documents/actor.mjs
```diff
@@ -516,39 +516,49 @@ export default class ActorEd extends Actor {
...
   /**
    * Only for actors of type Sentient (character, npc, creature, spirits, horror, dragon). Take the given amount of
    * damage according to the parameters.
    * @param {number} amount                                     The unaltered amount of damage this actor should take.
-   * @param {boolean} isStrain                                  Whether this damage is strain or not.
-   * @param {("standard"|"stun")} damageType                    The type of damage. One of either 'standard' or 'stun'.
-   * @param {("physical"|"mystical")} [armorType]               The type of armor that protects from this damage, one of either
+   * @param {object} [options]                                  The following options for taking damage:
+   * @param {boolean} [options.isStrain]                        Whether this damage is strain or not.
    *                                                            'physical', 'mystical', or 'none'.
```
ChristopherSD, line 540: It seems the rest of the documentation got lost with the PR merge? 😅

ChristopherSD (approved):

PR #1403, opened 2025-02-03
1086-chris-show-optional-devotions-on-questor-upgrade

### LINKED ISSUE #1086: ugrade a questor does not show the optional devotions to chose from.
upgrading a questor, does not show the prompt to chose an optional talent from the respective pool

ChristopherSD:
Fix #1086


patrickmohrmann:
something is really off with the advancement of questors. I get 3 times the dialog if i want to spend LP, only the second one is correct. I can create bug reports for it, there is a lot more broken and atleast now the optional talents are there.

patrickmohrmann (approved):

PR #1404, opened 2025-02-03
fixed trigger reaction story

### LINKED ISSUE #653: trigger reaction (button / reaction option prompt)
**Description**
- [x] the chat output of the attack will have a button **Reaction**. 
- [x] this button might be greyed out if no reactive ability is available at the target. (optional)
- [x] if the targets owner (or GM) clicks on the button, a prompt will appear with all defensive abilities listed as buttons. 
- [] ~tool tips of the button shall show the description of the ability~ (done in #954 )

**Additional context**

patrickmohrmann:
resolves #653

### DIFF
--- a/module/data/item/templates/targeting.mjs
+++ b/module/data/item/templates/targeting.mjs
@@ -53,7 +53,7 @@ export default class TargetTemplate extends SystemDataModel {
     let currentTarget = game.user.targets.first()?.actor;
     let currentTargets = [ ...game.user.targets.map( ( t ) => t.actor ) ];
     let numTargets = game.user.targets.size;
-    let targetDifficultySetting = this.difficulty.target;
+    let targetDifficultySetting = this.difficulty.target === "" ? 0 : this.difficulty.target;
     let groupDifficultySetting = this.difficulty.group;
     let fixedDifficultySetting = this.difficulty.fixed;
 
@@ -75,7 +75,7 @@ export default class TargetTemplate extends SystemDataModel {
           baseDifficulty = TargetTemplate._getAggregatedDefense( currentTargets, targetDifficultySetting, Math.min );
           break;
         default:
-          baseDifficulty = currentTarget?.system.characteristics.defenses[targetDifficultySetting].value ?? 0;
+          baseDifficulty = targetDifficultySetting === 0 ? 0 : currentTarget?.system.characteristics.defenses[targetDifficultySetting].value ?? 0;
       }
       difficulty = baseDifficulty + additionalTargetDifficulty;
     }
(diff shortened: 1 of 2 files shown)


module/data/item/templates/targeting.mjs
```diff
@@ -53,7 +53,7 @@ export default class TargetTemplate extends SystemDataModel {
     let currentTarget = game.user.targets.first()?.actor;
     let currentTargets = [ ...game.user.targets.map( ( t ) => t.actor ) ];
     let numTargets = game.user.targets.size;
-    let targetDifficultySetting = this.difficulty.target;
+    let targetDifficultySetting = this.difficulty.target === "" ? 0 : this.difficulty.target;
```
ChristopherSD, line 56: Can be simplified to 
`let targetDifficultySetting = this.difficulty.target || 0;`

Also, why is the default a `0`? This should be a valid key for the defenses, which 0 isn't. If you wanna say that we dont have that key, then semantically it would be better to use `undefined`, or in some cases `null`.
And this should be the case when it's not defined. This means we need to adjust the data model for that. It shouldn't be `blank`, but rather `required = false`.
If so, we don't need this variable at all.

module/data/item/templates/targeting.mjs
```diff
@@ -75,7 +75,7 @@ export default class TargetTemplate extends SystemDataModel {
           baseDifficulty = TargetTemplate._getAggregatedDefense( currentTargets, targetDifficultySetting, Math.min );
           break;
         default:
-          baseDifficulty = currentTarget?.system.characteristics.defenses[targetDifficultySetting].value ?? 0;
+          baseDifficulty = targetDifficultySetting === 0 ? 0 : currentTarget?.system.characteristics.defenses[targetDifficultySetting].value ?? 0;
```
ChristopherSD, line 78: See comments above about the `targetDifficultySetting` variable.
This line can be simplified either way:

ChristopherSD (changes requested):

module/data/item/templates/targeting.mjs
patrickmohrmann: ok

module/data/item/templates/targeting.mjs
patrickmohrmann: ok

ChristopherSD: This needs to be fixed. `targetDifficultySetting` can't be zero. It's  a string.
From what I understand you wanna achieve something like this?
```ecmascript 6
baseDifficulty = currentTarget?.system.characteristics.defenses[targetDifficultySetting]?.value ?? 0;
```

patrickmohrmann: ok, understood

ChristopherSD (approved):

PR #1408, opened 2025-02-05
automatically adding questor devotion during char gen

### LINKED ISSUE #1071: Add questor devotion to the actor on char gen
during character generation for questors, the **questor devotion** shall be added to the actor with rank 1 as well as it is done during adding a questor on an existing character.

patrickmohrmann:
resolves #1071


patrickmohrmann:
added core item and changed accordingly

ChristopherSD (changes requested):
The implementation is correct, but I found a bug that needs to be fixed so this PR can work. In `AbilityTemplate` the property `source` needs an initial value, otherwise, setting the source during character generation may throw an error. Putting this in `defineSchema` of `AbilityTemplate` works:
```javascript
source: new fields.SchemaField( {
        class: new fields.DocumentUUIDField( ClassTemplate, {
          label:    this.labelKey( "Ability.Source.class" ),
          hint:     this.hintKey( "Ability.Source.class" )
        } ),
        atLevel: new fields.NumberField( {
          required: false,
          nullable: true,
          min:      0,
          integer:  true,
          label:    this.labelKey( "Ability.Source.atLevel" ),
          hint:     this.hintKey( "Ability.Source.atLevel" )
        } ),
      },
      {
        required: false,
        initial:  {},
        label:    this.labelKey( "Ability.Source.class" ),
        hint:     this.hintKey( "Ability.Source.class" )
      } ),
```

ChristopherSD (approved):

PR #1412, opened 2025-02-06
Add ability as an edid field that can be searched for default edids from settings

ChristopherSD:
EdIdField seems to have some problem with initializing its defaults. The field here should be initialized with an empty string so it makes more sense for users to click into the input field.
This can be fixed later, it doesn't have implications for the functionality.

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -1,10 +1,13 @@
 {
   "ED": {
     "ActiveEffect": {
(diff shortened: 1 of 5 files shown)


patrickmohrmann (approved):

patrickmohrmann:
cool thats nice I  create a refactoring for the field default, but its great already

PR #1413, opened 2025-02-07
removed ed ids for know down and jump up and replaced it with actions

### LINKED ISSUE #1409: Refactor ed-ids into actiontypes
ed ids like jump up or knock down have to be added to actiontypes instead. ed ids will be needed for uniquly identifying abilities

patrickmohrmann:
resolves #1409

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -175,6 +175,11 @@ export default class ActorEd extends Actor {
     return itemTypes[type].filter( edidFilter );
   }
 
+  getAbilityByAction ( action ) {
+    const abilities = [ "talent", "skill", "devotion" ];
+    return this.items.filter( item => abilities.includes( item.type ) && item.system.rollType === action );
+  }
+
   /**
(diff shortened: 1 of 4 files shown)


module/documents/actor.mjs
```diff
@@ -175,6 +175,11 @@ export default class ActorEd extends Actor {
     return itemTypes[type].filter( edidFilter );
   }
 
+  getAbilityByAction ( action ) {
```
ChristopherSD, line 178: The name suggests that it returns exactly one Item. Filter returns an Array.
The type check does not add information. If the item doesn't have the property (which only rollables do), it doesn't return anything.
Since this is the replacement for searchin items by ed-id, it should probably look like this:
```ecmascript 6
getItemsByAction( action ) {
  return this.items.filter(
    item => item.system.rollType === action
  );
}
```

If it has to restrict to abilities, I would do that after the call of this function, so this function can remain more gerneric

ChristopherSD (changes requested):

module/documents/actor.mjs
patrickmohrmann: ok

ChristopherSD (approved):

PR #1438, opened 2025-02-12
1437 basic knack visualization

### LINKED ISSUE #1437: basic knack visualization
show knacks in the ui

patrickmohrmann:
resolves #1437


module/data/item/knack-ability.mjs
```diff
@@ -42,6 +42,26 @@ export default class KnackAbilityData extends AbilityTemplate.mixin(
...
+  
+
+  /**
+   * The final rank of the ability (e.g. attribute + rank).
+   * @type {number}
+   */
+  get rankFinal() {
+    const parentTalent = this.parentActor?.itemTypes.talent.find( ( talent ) => talent.system.edid === this.sourceTalent );
+    const attributeStep = ( this.parentActor?.system.attributes[this.attribute]?.step ?? 0 );
+    return attributeStep ? parentTalent?.system.level + attributeStep : 0;
```
ChristopherSD, line 62: I don't think you need the `parentTalent` here. You already got the `parentRank` getter, which is great. So you can just delete that line and replace this with `this.parentRank()`

module/data/item/knack-ability.mjs
```diff
@@ -42,6 +42,26 @@ export default class KnackAbilityData extends AbilityTemplate.mixin(
     // assign the source talent
   }
 
+  /**
+   * The final rank of the ability (e.g. attribute + rank).
```
ChristopherSD, line 46: Wrong documentation, it's the final rank of the parent talent

ChristopherSD (changes requested):
Nice!

I think the way the talents and knacks are rendered in "templates/actor/actor-tabs/talent.hbs" currently are not very performant.
For each category we iteratie over _all_ items, and for each of those items we iterate over _all_ knacks. I think this should be prepared in the sheet class and pass an object that we only have to iterate over once somehow

module/data/item/knack-ability.mjs
patrickmohrmann: I'll make a refactoring for your first comment #1456 

I think it is fair to say, it works like this for now but should be fixed before the final release.

and yes, great!

module/data/item/knack-ability.mjs
patrickmohrmann: ok

ChristopherSD (approved):

PR #1446, opened 2025-02-13
1359 power data model

### LINKED ISSUE #1359: Power Data Model
The Power data Model shall have the following Data fields 

- step
- damage step (optional)
- strain
- mixin rollable Template
  - rollType (everything except halfmagic, damage, effect, initiative from the config) -- initiative has to be checked if it is a thing... 
- mixin TargetTemplate
- opposedArmortype (mystical or physical) for powers with a damage step.

patrickmohrmann:
resolves #1359


module/data/item/power.mjs
```diff
@@ -1,16 +1,35 @@
...
+        initial:  "",
+        choices:  ED4E.armor,
+        label:    this.labelKey( "Weapons.armorType" ),
+        hint:     this.hintKey( "Weapons.armorType" )
+      } ),
+      damage:        new fields.SchemaField( {
+        type: new fields.StringField( {
+          initial:  "standard",
+          choices:  ED4E.damageType,
+          label:    this.labelKey( "Weapons.damageType" ),
```
ChristopherSD, line 29: Missing localization in de.json

module/data/item/power.mjs
```diff
@@ -1,16 +1,35 @@
...
+          hint:     this.hintKey( "Weapons.damageType" )
```
ChristopherSD, line 30: Missing localization in de.json

module/data/item/power.mjs
```diff
@@ -1,16 +1,35 @@
...
   static defineSchema() {
+    const fields = foundry.data.fields;
     return this.mergeSchema( super.defineSchema(), {
+      armorType: new fields.StringField( {
+        required: true,
+        nullable: true,
+        blank:    true,
+        initial:  "",
+        choices:  ED4E.armor,
+        label:    this.labelKey( "Weapons.armorType" ),
```
ChristopherSD, line 22: Wrong localization in de.json: Says "Waffe" instead of power

module/data/item/power.mjs
```diff
@@ -1,16 +1,35 @@
...
+        hint:     this.hintKey( "Weapons.armorType" )
```
ChristopherSD, line 23: Wrong localization in de.json: Says "Waffe" instead of power

ChristopherSD (changes requested):
Since you made the `damage` field a SchemaField anyway, an idea: how about adding a field for `element` and its subtype (like in spell)?
This way powers can be elemental?

module/data/item/power.mjs
patrickmohrmann: true

module/data/item/power.mjs
patrickmohrmann: true

module/data/item/power.mjs
patrickmohrmann: true

module/data/item/power.mjs
patrickmohrmann: true

module/data/item/power.mjs
```diff
@@ -1,16 +1,65 @@
...
+        } ),
+      } ),
+      element: new fields.SchemaField( {
+        type: new fields.StringField( {
+          required: true,
+          nullable: true,
+          blank:    true,
+          trim:     true,
+          choices:  ED4E.elements,
+          label:    this.labelKey( "Power.powerElementType" ),
```
ChristopherSD, line 40: I know this is nitpicking, but... 😅
I would just call it "elementType"
With the "Power" prefix we already know where we are, that's the idea. This way we can keep it "short", in heavy quotes :D

module/data/item/power.mjs
```diff
@@ -1,16 +1,65 @@
...
+          required: true,
+          nullable: true,
+          blank:    true,
+          trim:     true,
+          choices:  Object.values(
+            ED4E.elementSubtypes
+          ).map(
+            subtypes => Object.keys( subtypes )
+          ).flat(),
+          label:    this.labelKey( "Power.powerElementSubtype" ),
```
ChristopherSD, line 53: See comment on damage element localization

module/data/item/power.mjs
patrickmohrmann: you are right, power is already done by the Power. grouping

module/data/item/power.mjs
patrickmohrmann: ok

ChristopherSD (approved):

PR #1454, opened 2025-02-14
add validation for namegiver items if create on actors

### LINKED ISSUE #1080: Allow only one namegiver per Actor
**Description**
NPCs and character actors shall only be allowed to have one namegiver item 

**Additional context**

patrickmohrmann:
resolves #1080

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -1795,7 +1795,9 @@
       "windling":                                         "Windling"
     },
     "Notifications": {
-      "Error": {},
+      "Error": {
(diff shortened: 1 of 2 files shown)


ChristopherSD (commented):
This is already prepared to be implemented in a different way. The `metadata` property of  `ItemDataModel` has the `singleton` field, which indicates if a data model/Item can only exist once in a parent. So what needs to be done, is to set this field to true in the `NamegiverData` and implement the handling. 
You can have a look at it in the dnd5e system on the 4.2.x branch.
Let's do that together!

PR #1455, opened 2025-02-14
make movement data model nullable

### LINKED ISSUE #1037: Namegiver movement type shall be nullable
a movement type of 0 is not the correct way of showing that the Namegiver does not has any special movement method. it should be null for **no special method** and 0 for **no movement possible for this type**

this topic has to be looked at when we are working on tokens and canvas

patrickmohrmann:
resolves #1037


ChristopherSD:
@patrickmohrmann did you check what the fields are initialized with? Do they have a correct initial value of `null`?
The methods for initializing have changed a bit in V13, just need to make sure...
I don't have a computer test currently...

ChristopherSD (approved):

PR #1457, opened 2025-02-18
1448 rework attribute roll to have pre defined option

### LINKED ISSUE #1448: rework attribute roll to have pre-defined option
triggering an attribute roll shall create a small prompt before the roll prompt appears.

refactor it also similar to the workflow refactoring see PR #1621 

this prompt shall contain the following options:

| Half magic | Attribute Roll | Substitute Skill |
|------------|----------------|------------------|
| Discipline 1 | --- | avoid blow |
| Discipline 2 | --- | climb |
| --- | --- | jump |

add all attack options (melee, unamred etc. to dextertity as well as substitude skills. Add tail attack as well.

patrickmohrmann:
resolves #1448 

halfmagic and substitute roll is currently only a notification,

### DIFF
--- a/earthdawn4e.css
+++ b/earthdawn4e.css
@@ -75,6 +75,9 @@ input.input__number-input {
   padding-top: 5px;
   height: 20px;
 }
+.no-footer .form-footer {
+  display: none;
+}
 .one-line__vertical-align > * {
   line-height: 35px;
   vertical-align: middle;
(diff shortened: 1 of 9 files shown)


ChristopherSD (commented):
We should go over this together, there are several points to clear up, like:
- hard coding the abilities in the config (this needs to be dynamic somehow, if necessary at all)
- the multiple prompt types for each attribute (if we need the factory to accept input params, like the attribute, we need to think about if that's what the factory should do in the first place, and, if so, refactor it first)
- the handling of the roll shouldn't happen all in `rollAttribute`, this should only "roll an attribute", there should be handling in the sheet and corresponding delegating to the responsible functions
- the styling for only showing the buttons on hover (this is possibly a programmatic approach, therefore should already be included here)

PR #1465, opened 2025-02-25
1366-extend-getrolldata-method-for-actor-and-item-documents

### LINKED ISSUE #1366: Extend `getRollData` method for Actor and Item documents
## Story

As a player/GM, I want the simplest syntax possible for `FormulaField`s, so I don't need to remember complex system field paths, and can easily use that feature. 

## detail description

To simplify the syntax for `FormulaField`s, e.g. not having to write something like `2 + @system.characteristics.health.damage.total` but rather `2 + @damage`.
Formula fields will be evaluated, like Rolls, against provided data to allow for variable calculation. This can include values like the rank of an ability, or the Dex step of a character. For `Actor`s and `Item`s, it is usually their system data.  Often the different properties have a too complex field path in the system data. It's necessary for a good UX and quality of life to simplify this.

## Proposed Solution

This data is based on their `getRollData` method, which only provides the system data in its base version. This method can be extended to provide shortcuts to often used values.

## Acceptance Critera
At least the following shortcut properties should be provided:

- [x] attribute steps 
    - `@dex`
    - `@str`
- [x] armor values
    - `@pa`
    - `@ma`
- [x] defense values
    - `@pd`
    - `@md`
    - `@sd`
- [x] health values
    - `@death`
    - `@unconscious`
    - `@woundThresh`
    - `@wounds`
    - `@damage`
- [x] movement values
    - `@burrow`
    - `@climb`
    - `@fly`
    - `@swim`
    - `@walk`
- [x] resource steps
    - `@karma`
    - `@devotion`
    - `@recovery`
- [x] initiative step
    - `@ini` 
- ... tbd

ChristopherSD:
Closes #1366


module/config.mjs
patrickmohrmann, line 1: is there a reason to keep this?

patrickmohrmann (approved):

PR #1466, opened 2025-02-25
444 show karma and maneuver knacks

### LINKED ISSUE #444: show karma and maneuver knacks
include karma and maneuver knacks as separate entries in the talent overview

place some icons in front of each

✨ fa-sparkles – Karma Knack
🌀 fa-hurricane – Special Maneuver
⚡ fa-bolt – knack abilities

patrickmohrmann:
in short, this is what this pr is about. I added all kinds of knack visualizations

resolves #444 

image

### DIFF
--- a/module/applications/item/class-item-sheet.mjs
+++ b/module/applications/item/class-item-sheet.mjs
@@ -162,15 +162,6 @@ export default class ClassItemSheetEd extends ItemSheetEd {
       }
     );
 
-    
-    context.enrichedBriefDescription = await TextEditor.enrichHTML(
-      this.document.system.briefDescription.value,
-      {
-        // Only show secret blocks to owner
-        secrets:    this.document.isOwner,
-      }
-    );
-
     return context;
   }
 
--- a/module/applications/item/class-item-sheet.mjs
+++ b/module/applications/item/class-item-sheet.mjs
@@ -162,6 +162,15 @@ export default class ClassItemSheetEd extends ItemSheetEd {
       }
     );
 
+    
+    context.enrichedBriefDescription = await TextEditor.enrichHTML(
+      this.document.system.briefDescription.value,
+      {
+        // Only show secret blocks to owner
+        secrets:    this.document.isOwner,
+      }
+    );
+
     return context;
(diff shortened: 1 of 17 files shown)


module/applications/item/class-item-sheet.mjs
```diff
@@ -162,6 +162,15 @@ export default class ClassItemSheetEd extends ItemSheetEd {
       }
     );
 
+    
+    context.enrichedBriefDescription = await TextEditor.enrichHTML(
```
ChristopherSD, line 166: Why do we need this here when its already in `ItemSheetEd`? We're calling super, so its already there.

module/applications/item/physical-item-sheet.mjs
```diff
@@ -109,6 +109,14 @@ export default class PhysicalItemSheetEd extends ItemSheetEd {
         // rollData: this.document.getRollData
       }
     );
+
+    context.enrichedBriefDescription = await TextEditor.enrichHTML(
```
ChristopherSD, line 113: Why do we need this here when its already in `ItemSheetEd`? We're calling super, so its already there.

module/data/item/templates/item-description.mjs
```diff
@@ -19,6 +19,14 @@ export default class ItemDescriptionTemplate extends SystemDataModel {
           hint:     this.hintKey( "description" ),
         } ), 
       } ),
+      briefDescription: new fields.SchemaField( {
```
ChristopherSD, line 22: naming suggestion: `summary`
so there's no overlap with `description`. makes it clearer?

ChristopherSD (changes requested):

module/applications/item/physical-item-sheet.mjs
patrickmohrmann: true

module/data/item/templates/item-description.mjs
patrickmohrmann: good idea

ChristopherSD (approved):

PR #1483, opened 2025-03-06
1480 smaller lp tracking fixes

### LINKED ISSUE #88: [LP Tracking] Knacks
Knacks can be learned by drag and drop it to the actor.

### LINKED ISSUE #1226: [BUG]-Class at rank 0 after character generation
the class is at rank 0 after character generation. it has to be at rank 1 of course. I believe the issue is based on the rank of the class item. All Class items shall have the base value of 0 in the compendium, but shall be added as class level 1.

patrickmohrmann:
includes knack learn prompt as well as other smaller changes
resolves #88 
resolves #1226


module/applications/global/prompt-factory.mjs
```diff
@@ -379,6 +379,7 @@ class ItemPromptFactory extends PromptFactory {
 
   _promptTypeMapping = {
     chooseTier:     this._chooseTierPrompt.bind( this ),
+    learnKnack:     this._learnKnackPrompt.bind( this ),
```
ChristopherSD, line 382: Why a special prompt for that? It's an ability, so it could use the ability prompt?

module/applications/actor/character-sheet.mjs
```diff
@@ -237,9 +237,8 @@ export default class ActorSheetEdCharacter extends ActorSheetEdNamegiver {
 
   /** @inheritdoc */
   async _onDropItem( event, item ) {
-    if ( item.system.learnable ) item.system.constructor.learn( this.actor, item );
+    if ( item.system.learnable ) return item.system.constructor.learn( this.actor, item );
```
ChristopherSD, line 240: why return it here?
if needed, the second line needs a return as well. and the documentation has to be adjusted (it currently says it returns a `Promise<void>`)

module/applications/global/prompt-factory.mjs
patrickmohrmann: I won :-)

ChristopherSD (approved):

PR #1492, opened 2025-03-08
1360 roll powers

### LINKED ISSUE #1360: Roll powers
powers are similar to talents except that they do not use an attribute, but instead roll the flat step. The Rollable Template shall be enhanced to take the step of the ability instead of Attribute+Rank

### LINKED ISSUE #1361: Power attacks
powers with the ed-id creature-attack shall work similar to a weapon attack but:
- do not require a weapon (and equipment status etc.)
- use the damage step from the power itself

patrickmohrmann:
resolves #1360 
resolves #1361

### DIFF
--- a/templates/item/item-partials/item-details/details/item-details-power.hbs
+++ b/templates/item/item-partials/item-details/details/item-details-power.hbs
@@ -9,3 +9,19 @@
     {{formField systemFields.element.fields.subtype name="system.element.subtype" value=item.system.element.subtype localize=true}}
 </fieldset>
 {{> "systems/ed4e/templates/item/item-partials/item-details/partials/targeting.hbs"}}
+
+<fieldset>
+  <legend>{{localize "ED.Item.Header.actionType"}}</legend>
+
+  {{!-- abilty action type --}}
+  {{formGroup systemFields.rollType name="system.rollType" value=item.system.rollType localize=true}}
+
+  {{#each systemFields.rollTypeDetails.fields as |field key|}}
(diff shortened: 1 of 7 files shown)


templates/item/item-partials/item-details/details/item-details-power.hbs
```diff
@@ -8,4 +8,20 @@
     {{formField systemFields.element.fields.type name="system.element.type" value=item.system.element.type localize=true}}
     {{formField systemFields.element.fields.subtype name="system.element.subtype" value=item.system.element.subtype localize=true}}
 </fieldset>
-{{> "systems/ed4e/templates/item/item-partials/item-details/partials/targeting.hbs"}}
\ No newline at end of file
+{{> "systems/ed4e/templates/item/item-partials/item-details/partials/targeting.hbs"}}
+
+<fieldset>
```
ChristopherSD, line 13: This should go to the very top, like in the other items. It's probably the most important field

ChristopherSD (changes requested):

templates/item/item-partials/item-details/details/item-details-power.hbs
ChristopherSD, line 13: Also, this is used multiple times now and is quite complex. So it should have its own partial, imho

patrickmohrmann: ok

patrickmohrmann: ok

PR #1496, opened 2025-03-10
1495 smaller changes to current sheet to fulfill creature needs

### LINKED ISSUE #1495: Smaller changes to current sheet to fulfill creature needs
remove unnecessary features from the creature sheet.

patrickmohrmann:
resolves #1495 

finished up the creature sheet by removing all none creature related options

### DIFF
--- a/module/data/actor/templates/sentient.mjs
+++ b/module/data/actor/templates/sentient.mjs
@@ -279,6 +279,16 @@ export default class SentientTemplate extends CommonTemplate {
         label: this.labelKey( "challenge" ),
         hint:  this.hintKey( "challenge" ),
       } ),
+      actions: new fields.NumberField( {
+        required: true,
+        nullable: false,
+        min:      1,
+        step:     1,
+        initial:  1,
+        integer:  true,
+        label:    this.labelKey( "actions" ),
+        hint:     this.hintKey( "actions" ),
+      } ),
       condition: new fields.SchemaField( {
         aggressiveAttack: new fields.BooleanField( {
           required: true,
--- a/module/data/actor/templates/sentient.mjs
+++ b/module/data/actor/templates/sentient.mjs
@@ -258,37 +258,6 @@ export default class SentientTemplate extends CommonTemplate {
         label: this.labelKey( "characteristics" ),
         hint:  this.hintKey( "characteristics" ),
       } ),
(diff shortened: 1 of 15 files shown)


module/data/actor/templates/sentient.mjs
```diff
@@ -279,6 +279,16 @@ export default class SentientTemplate extends CommonTemplate {
         label: this.labelKey( "challenge" ),
         hint:  this.hintKey( "challenge" ),
       } ),
+      actions: new fields.NumberField( {
```
ChristopherSD, line 282: mhh, not super happy that this is in "sentient" since this applies to PCs then as well. So either find a way to only apply it NPCs (like a NPC template to mix in), or make it nullable, so it's clear that PCs don't use that. I'd prefer the first one, that makes it more correct and extendable

templates/actor/actor-tabs/equipment.hbs
```diff
@@ -1,24 +1,29 @@
 <section class="tab {{tabs.equipment.cssClass}}" data-group="sheet" data-tab="equipment">
-{{localize "ED.Actor.Header.carryingCapacity"}}
+    {{log this}}
```
ChristopherSD, line 2: No logging pleeeeaaase :))

templates/actor/actor-tabs/equipment.hbs
```diff
@@ -1,24 +1,29 @@
...
+    {{#unless (ne this.actor.type "character")}}
```
ChristopherSD, line 3: this basically is `if type === 'character'` ? :D

ChristopherSD (changes requested):

module/data/actor/templates/sentient.mjs
patrickmohrmann: done

templates/actor/actor-tabs/equipment.hbs
patrickmohrmann: yes okay :-P

templates/actor/actor-tabs/equipment.hbs
patrickmohrmann: ok

ChristopherSD (approved):

PR #1502, opened 2025-03-13
1501 bug talent ranks are not reset on changing class

### LINKED ISSUE #975: changing optional talents will not reset spent points
switching the optional talent during character generation, after applying ranks, will not reset the ranks applied to the optional talent. 

**workaround** reset all talent ranks

### LINKED ISSUE #1048: [BUG]- create character without char gen
it shall be possible to create an actor of type character without character generation, currently it is not possible

### LINKED ISSUE #1501: [BUG]-talent ranks are not reset on changing class
changing the class does not reset the talent ranks

patrickmohrmann:
resolves #1501 reset talent points on class change
resolves #975  add spend optional points back to available points after changing an optional talent
resolves #1048 create character without char gen

### DIFF
--- a/module/applications/actor/character-generation-prompt.mjs
+++ b/module/applications/actor/character-generation-prompt.mjs
@@ -591,13 +591,6 @@ export default class CharacterGenerationPrompt extends HandlebarsApplicationMixi
   
   static _onReset( _, target ) {
     const resetType = target.dataset.resetType;
-    // if ( this.charGenData === undefined ) {
-    //   await charGenData.resetPoints( resetType );
-    //   instance.render();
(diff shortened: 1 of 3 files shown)


ChristopherSD (commented):
Let's please have a look at it together. I'm not sure about the changes in the `_onReset` of the char gen prompt.
Also: we need to discuss the changes to document creation. We already have a setting, if we always ask if it should be opened, that is superfluous. I thought the idea was, to introduce an alternate click (e.g. shift+click) to avoid the char gen dialog).

ChristopherSD (approved):

PR #1504, opened 2025-03-14
recover from stun only if stun damage is >0

### LINKED ISSUE #870: recover stun damage should not be an option if recovery tests value are 0
recover stun damage should not be an option if recovery tests value are 0

patrickmohrmann:
resolves #870

### DIFF
--- a/module/applications/global/prompt-factory.mjs
+++ b/module/applications/global/prompt-factory.mjs
@@ -163,7 +163,8 @@ class ActorPromptFactory extends PromptFactory {
       class:   "recovery default button-recovery",
       default: false
     } );
-    if ( this.document.system.characteristics.recoveryTestsResource.stunRecoveryAvailable && this.document.system.characteristics.health.damage.stun > 0 ) buttons.push( {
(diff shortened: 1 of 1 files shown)


module/applications/global/prompt-factory.mjs
```diff
@@ -163,7 +163,7 @@ class ActorPromptFactory extends PromptFactory {
       class:   "recovery default button-recovery",
       default: false
     } );
-    if ( this.document.system.characteristics.recoveryTestsResource.stunRecoveryAvailable ) buttons.push( {
+    if ( this.document.system.characteristics.recoveryTestsResource.stunRecoveryAvailable && this.document.system.characteristics.health.damage.stun > 0 ) buttons.push( {
```
ChristopherSD, line 166: line is too long. please split before the `&&` :)

ChristopherSD (changes requested):

module/applications/global/prompt-factory.mjs
patrickmohrmann: ok

ChristopherSD (approved):

PR #1505, opened 2025-03-14
fix wound calculation for strain

### LINKED ISSUE #972: Strain shall never cause a wound
**Describe the bug**
adding a high amount of strain to an ability and executing it. a wound is created if woundthreshold is below the strain. This shall not happen

patrickmohrmann:
resolves #972 

strain never causes a wound and therefore never causes knockdown

in addition, a wound from standard damage is created if the damage is **equal** or higher

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -632,7 +632,7 @@ export default class ActorEd extends Actor {
 
     const updates = { [`system.characteristics.health.damage.${ damageType }`]: newDamage };
 
-    if ( damageTaken >= health.woundThreshold && options.isStrain !== true ) {
+    if ( damageTaken >= health.woundThreshold && !options.isStrain ) {
       switch ( damageType ) {
         case "standard":
           updates["system.characteristics.health.wounds"] = health.wounds + 1;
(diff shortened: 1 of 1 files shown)


module/documents/actor.mjs
```diff
@@ -632,7 +632,7 @@ export default class ActorEd extends Actor {
 
     const updates = { [`system.characteristics.health.damage.${ damageType }`]: newDamage };
 
-    if ( damageTaken > health.woundThreshold ) {
+    if ( damageTaken >= health.woundThreshold && options.isStrain !== true ) {
```
ChristopherSD, line 635: can be written as `&& !options.isStrain`

module/documents/actor.mjs
```diff
@@ -662,7 +662,7 @@ export default class ActorEd extends Actor {
       ChatMessage.create( messageData );
     }
 
-    const knockdownTest = !this.system.condition.knockedDown && damageTaken >= health.woundThreshold + 5;
+    const knockdownTest = !this.system.condition.knockedDown && damageTaken >= health.woundThreshold + 5 && options.isStrain !== true;
```
ChristopherSD, line 665: can be written as `&& !options.isStrain`

ChristopherSD (changes requested):

module/documents/actor.mjs
patrickmohrmann: done

module/documents/actor.mjs
patrickmohrmann: done

ChristopherSD (approved):

PR #1519, opened 2025-03-15
add spellcasting option to discipline sheet

### LINKED ISSUE #1517: add spellcasting option to discipline sheet
add the spellcasting option to the discipline sheet

patrickmohrmann:
resolves #1517

### DIFF
--- a/module/applications/advancement/class-advancement.mjs
+++ b/module/applications/advancement/class-advancement.mjs
@@ -50,6 +50,7 @@ export default class ClassAdvancementDialog extends HandlebarsApplicationMixin( 
     this.effectsGained = this.classItem.system.advancement.levels[ this.nextLevel - 1 ].effects;
   }
 
+
   /**
    * Wait for dialog to be resolved.
    * @param {ItemEd} classItem                  The class item for which to display advancement options.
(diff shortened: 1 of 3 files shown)


ChristopherSD (commented):
I'm not sure: do we actually still need that field at all? 😅
What do we do with it?

--> make sense to check for the thread weaving talent, since we have to filter for the type.

patrickmohrmann:
I did something and it is working, please check the last commit

module/applications/advancement/class-advancement.mjs
```diff
@@ -155,6 +156,37 @@ export default class ClassAdvancementDialog extends HandlebarsApplicationMixin(
     return this.close();
   }
 
+  /**
+   * Get the class casting type for the given class item.
+   * @param {ItemEd} classItem - The class item to get the casting type for.
+   * @returns {Promise<string|null>} - The class casting type, or null if not found.
+   * @userFunction UF_ClassAdvancementDialog_getClassCastingType
+   */
+  static async getClassCastingType( classItem ) {
```
ChristopherSD, line 165: this should be an instance method of the class item itself, that is, the discipline data model.

ChristopherSD (commented):
Seeing the effort it needs to find the spellcasting item, and theoretically, the uncertainty on the result: I'm not a 100% convinced we should actually do that. 
I'm so sorry 😟
Can we have another live discussion about it?

PR #1530, opened 2025-03-23
Migration 1 test

patrickmohrmann:
please ignore a lot of those files, the important ones to get the idea of what I tried to accomplish are 

Documents/item-migration/talant-migration.mjs
Data/item/talent.mjs

everything else is more or less a spin of of them


patrickmohrmann:
keep it as backup for the moment

PR #1555, opened 2025-03-28
CSS and Accompanying EN Lang changes

alegionofone:
This is the first large change to the css code. I had added my own code to the end of the css for now with only minimal changes made to the main body of the code (one or two minor ones that I didn't want to inherit and couldn't override). This is easily overridden simply by deleting all code that appears after /* Alex's Alterations */ (with the odd change, as mentioned, such as a height: that I had to alter in the original code)

I have also edited the EN lang file a little more, firstly so that some labels looked better on the sheet and remove superfluous words and secondly some I missed on the first pass or additions.


alegionofone:
I just realised I pushed .png files, I would be converting these to .webp at some point for reduced memory. But lets run with this for now.

PR #1559, opened 2025-03-29
1515-multi-level-conditions

### LINKED ISSUE #1515: multi level conditions
Conditions: multiple levels (like Harried), examples in Thell's Artichron System or dnd5e

ChristopherSD:
Resolves #1515

All taken from system "Artichron" from Zhell (Discord), thank you soooooo much!!!!

Conditions can now have multiple levels. This is configurable. They can currently only be assigned via TokenHUB since they are a custom document subtype. There you can increase the level with left click, and decrease with right click.


module/config/statuses.mjs
```diff
@@ -132,6 +133,8 @@ export const statusEffects = [
     name: "ED.ActiveEffect.Status.aggressive",
     img:  "systems/ed4e/assets/icons/confrontation.svg",
 
+    reference: "TODO: Compendium UUID to explanation",
```
patrickmohrmann, line 136: If i understand this correct, it is a link to the Journal where the status is explained, if so, can we add multiple Uuids for german and english and maybe even polish and french?

ChristopherSD: Hm, right... Yeah we can do that. Could we open a new issue for that? So we can do it, when the compendia are available/finished?

ChristopherSD:
Yes. I made Harried a normal on/off condition and left Overwhelmed as multiple levels. There's no automation yet, but it is correctly implemented with Harried giving -2 and each instance of Overwhelmed -1

module/config/statuses.mjs
patrickmohrmann: done #1569

patrickmohrmann:
but you can add multiple harried conditions similar to overwhelmed

ChristopherSD:
Well, I think this is up to us how to do it. This the rules text:
> Multiple sources of distraction could further disorient an already harried character.
A harried character suffers a further –1 penalty to his tests and Defense ratings for
each additional harried effect he incurs after the first.

This means, you can have multiple harried effects, each above 1 add -1. And when there's more than 1 harried it effect, the whole bunch of them is called "overwhelmed". That's how I understand it.
So, it's either Harried fixed and multiple overwhelmed. Or multiple harried and overwhelmed fixed. I chose the former bc this way both conditions actually have an effect and are not just an image. Also, the way conditions with levels are incremented currently just multiplies their base effect. This wouldn't work if Harried was the one with multiple levels, since it would always give an additional -2 penalty.

Would that be okay as an explanation for now, and maybe we can wait and see if users give feedback on this themselves? (Just because it's a bit simpler implementation-wise for now)

PR #1560, opened 2025-03-29
Enhance delete functionality to support direct deletion with Shift key

ChristopherSD:
Works for Actor Sheets

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -1,2032 +1,2022 @@
 {
-  "ED": {
-    "ActiveEffect": {
-      "newEffectName": "Neuer Earthdawn ActiveEffect",
-      "ChangeKeys": {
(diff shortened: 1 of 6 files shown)


patrickmohrmann:
oh this is super dangerous don't you thing, I see the benefit from it, but I would at least have a system setting protecting this function.

ChristopherSD:
We don't have to do it. Was just getting annoyed bc I had to delete so many effects for testing 😅

patrickmohrmann:
I see the benefit for it doing development and session prep but i find it dangerous in game play. 

so I would vote for a setting,

ChristopherSD:
Okay, Implemented it

patrickmohrmann (approved):

PR #1566, opened 2025-03-30
Global .less + unused .less moved to old-less folder

alegionofone:
I have moved all files, declared unused by the earthdawn4e.less file, to a folder called old-less. This keeps them present but out of the way and helps me understand which I need to be working on.

I have made a few global changes to the .less file:

Window header formatting is complete.
A couple of global changes made, such as removal of backdrop-filters on actors from core css. I have included the character sheet images and done some minor work to the less to prove the backgrounds function with the new headers.


patrickmohrmann:
on the right side is the light sheet, this is not usable 
image

less/global/global.less
```diff
@@ -5,27 +5,39 @@
...
 .partial__six {
     float: left;
     width: 16.6%;
 }
-
+//###########################################################
+//                       window header
+//###########################################################
+.window-header {
+    background: none;
```
patrickmohrmann, line 28: Leaving the header without any background creates an empty space, I would suggest to put the name into the header of the actor (or move it up).

In addition, since the header is used for dragging the sheet around, i would find it helpful to have at least a very low opaque visual indicator that there is a header bar.

right now it looks somehow lost or that something should be there 
image

patrickmohrmann:
light mode not done yet. see https://github.com/patrickmohrmann/earthdawn4eV2/issues/1573

alegionofone:
I am in the process of breaking this up into a few smaller changes, en.json has already been done and I have pushed the .less file archiving for approval too. I will start back on the .less formatting now, but I would like to try to work out why I wasn't getting the same error that you were above, as it'd probably better to have that worked out before I continue too far as it may reoccur.

I will delete this when I have removed all changes I want from it.

PR #1577, opened 2025-04-02
Migration data model basics

patrickmohrmann:
First migration PR, as agreed only the document migration to differentiate the old to the new item types

### DIFF
--- a/module/documents/item.mjs
+++ b/module/documents/item.mjs
@@ -1,20 +1,6 @@
 import DocumentCreateDialog from "../applications/global/document-creation.mjs";
 import AdvancementLevelData from "../data/advancement/advancement-level.mjs";
-import * as game from "../hooks/_module.mjs";
-import ArmorMigration from "./migration/item/old-system/armor.mjs";
-import DevotionMigration from "./migration/item/old-system/devotion.mjs";
-import DisciplineMigration from "./migration/item/old-system/discipline.mjs";
-import EquipmentnMigration from "./migration/item/old-system/equipment.mjs";
-import KnackMigration from "./migration/item/old-system/knack.mjs";
-import MaskMigration from "./migration/item/old-system/mask.mjs";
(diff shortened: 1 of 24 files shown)


module/documents/migration/actor/old-system/none-character.mjs
```diff
@@ -0,0 +1,17 @@
+export default class NoneCharacterMigration {
+
+  static async migrateData( source ) {
+
+    // define new actor type by system.actorType
+    if ( source.system.actorType ) { 
+      if ( source.system.actorType === "creature" ) source.type = "creature";
```
ChristopherSD, line 7: I'd say that is a whole lot of `if`/`else`. How about a "migration.mjs" in the config ?

module/documents/item.mjs
```diff
@@ -252,4 +266,28 @@ export default class ItemEd extends Item {
...
   // endregion
 
+  /* -------------------------------------------- */
+  /*  Migrations                                  */
+  /* -------------------------------------------- */
+
+  static migrateData( source ) {
+    source = super.migrateData( source );
+
+    if ( source.type === "armor" )            ArmorMigration.migrateData( source );
```
ChristopherSD, line 276: Same as with actors: a whole lot of `if`/`else`
Make a config thing out it. That way we don't need to Import that much either...

ChristopherSD (changes requested):

module/documents/migration/actor/old-system/none-character.mjs
patrickmohrmann: not sure how i feel about it, its not something used somewhere else but here I believe

module/documents/item.mjs
patrickmohrmann: don't get it, than i have to import it all in the new config don't I??? something for the next time maybe

ChristopherSD (approved):

PR #1635, opened 2025-04-17
finish canvas documentation

patrickmohrmann:


module/canvas/token.mjs
```diff
@@ -89,6 +90,14 @@ export default class TokenEd extends Token {
     this._setBarPosition( bar, number, barHeight );
   }
 
+  /**
+   * Reset the bar graphics to a default state.
+   * @param {object} bar     PIXI.Graphics instance for the bar
```
ChristopherSD, line 95: Make the type `PIXI.Graphics` if you know that that's it

module/canvas/token.mjs
```diff
@@ -97,6 +106,13 @@ export default class TokenEd extends Token {
       .drawRoundedRect( 0, 0, width, height, 3 );
   }
 
+  /**
+   * Set the position of the bar based on its order.
+   * @param {object} bar      PIXI.Graphics instance for the bar
```
ChristopherSD, line 111: Make the type `PIXI.Graphics` if you know that that's it

ChristopherSD (changes requested):

module/canvas/token.mjs
patrickmohrmann: please check if it is meant like this.

module/canvas/token.mjs
patrickmohrmann: please check if it is meant like this.

module/canvas/token.mjs
ChristopherSD: It's correct like in the doc for `_drawKarmaBar`, but I don't see changes here?

module/canvas/token.mjs
ChristopherSD: It's correct like in the doc for `_drawKarmaBar`, but I don't see changes here?

ChristopherSD (changes requested):

module/canvas/token.mjs
patrickmohrmann: ok

module/canvas/token.mjs
patrickmohrmann: ok

ChristopherSD (approved):

PR #1637, opened 2025-04-18
Actor Sheet Re-Style

alegionofone:
This is a big commit and involves .less and .hbs for character sheet styling. I have also made some changes to the en.json as I worked.

This is still something of a draft but I have made some alterations here that will knock on to sheets I have not refined: mostly going from partial__x to flex which will effect item sheets (these are still very draft so not a huge issue).


alegionofone:
This does not include the Light formatting I should add, some minor work has been done to make it readable, but not near complete. I think I will move away from the parchment, but I will keep the file there for future parchment use.

alegionofone:
Light mode has not been completed, it is currently made functional but by no means even a rough draft. I will move away from the parchment I think, but keep the file for future parchment use.

patrickmohrmann (approved):

PR #1638, opened 2025-04-19
Add localization JSON consolidation and linting utilities

ChristopherSD:
The pre-commit hook now fails if es-lint finds non-fixable issues.

JSON lang linting has been extended. When a lang file has been changed, all lang files are consolidated. This means, that all lang files are updated to have all the same keys across all files. If a key is inserted the get a default "TODO: Add translation" value.
This needs to be done once, to bring all lang files up to the same state, and from then on, changes in PRs in only one lang file are reflected in other files as well.

### DIFF
--- a/.lintstagedrc.yaml
+++ b/.lintstagedrc.yaml
@@ -1,8 +1,14 @@
 # lint and fix modules
-"module/**/*.mjs": "eslint --fix --config eslint.config.mjs"
 
-# format and check language files
-"lang/**/*.json": "npm run format:lang"
+"{module,utils}/**/*.mjs":
+  # fixable rules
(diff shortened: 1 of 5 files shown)


patrickmohrmann:
I'm not getting what this does????

ChristopherSD:
In the pre-commit hook, if any number of json files in the lang directory (so the localizations) have changed for the commit:
it takes all localization files, compares them, and brings them all to a version, where all of them have all the same keys. That means, it inserts keys in files that are present in others, but not in this one.

Does that make sense?

patrickmohrmann:
Thats what I thought it would, but it does not do it right now thats. what puzzled me

patrickmohrmann (approved):

PR #1640, opened 2025-04-20
matrix-data-model

ChristopherSD:
Create a matrix data model that can be embedded in other items.
This is the foundation, additional functionality will come with later stories.


patrickmohrmann (approved):

patrickmohrmann (commented):
ok

PR #1647, opened 2025-04-24
Move ability magic details into roll type

ChristopherSD:
Since "thread weaving" and "spellcasting" roll types are now part of abilities, the information in the magic field can be moved there. The castingType is a roll type detail for "thread weaving". Naming is getting confusing, but that's just the nature of spellcasting (which is one talent) vs. thread weaving (multiple talents), but only casting disciplines are different.

This also includes getting a potential casting type from a class item. May be relevant to PR #1519

### DIFF
--- a/module/data/item/templates/ability.mjs
+++ b/module/data/item/templates/ability.mjs
@@ -122,8 +122,6 @@ export default class AbilityTemplate extends ActionTemplate.mixin(
             blank:    false,
             trim:     true,
             choices:  ED4E.spellcastingTypes,
-            label:    this.labelKey( "Ability.Magic.magicType" ),
-            hint:     this.hintKey( "Ability.Magic.magicType" )
           } ),
         }, {} ),
       }, {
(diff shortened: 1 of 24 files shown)


module/data/item/templates/ability.mjs
```diff
@@ -111,15 +110,21 @@ export default class AbilityTemplate extends ActionTemplate.mixin(
...
         }, {
-          required: false,
           label:    this.labelKey( "Ability.RollTypeDetails.reaction" ),
           hint:     this.hintKey( "Ability.RollTypeDetails.reaction" ),
         } ),
         recovery:      new fields.SchemaField( {}, {} ),
         spellcasting:  new fields.SchemaField( {}, {} ),
-        threadWeaving: new fields.SchemaField( {}, {} ),
+        threadWeaving: new fields.SchemaField( {
+          castingType: new fields.StringField( {
```
patrickmohrmann, line 119: I missed it in one of the previous PRs, are you sure in adding the talent specific options into the ability template? why?

patrickmohrmann (approved):

PR #1650, opened 2025-04-25
Damage workflow

patrickmohrmann:
I created a PR to ensure its not going to be forgotten, please have alook if this - even if it works - is correct

### DIFF
--- a/module/workflows/workflow/damage-workflow.mjs
+++ b/module/workflows/workflow/damage-workflow.mjs
@@ -109,12 +109,28 @@ export default class DamageWorkflow extends ActorWorkflow {
     this._damageItem = this._options.damageItem ?? this._actor.attributes.str.step;
   }
 
+  // async #createRollOptions() {
+  //   this._rollOptions = DamageRollOptions.fromActor(
+  //     {
+  //       ...this.getCommonDamageRollData(),
+  //       damageItem: this._damageItem?.uuid ?? null,
+  //       chatFlavor: game.i18n.format( "TODO.ED.Chat.Flavor.Damage", {} )
+  //     },
+  //     this._actor,
+  //   );
+  // }
+
   async #createRollOptions() {
+    const commonRollData = await this.getCommonDamageRollData();
     this._rollOptions = DamageRollOptions.fromActor(
       {
-        ...this.getCommonDamageRollData(),
+        ...commonRollData,
+        step: {
+          ...commonRollData.step,
+          base: await this.calculateDamageStep(), 
+        },
         damageItem: this._damageItem?.uuid ?? null,
-        chatFlavor: game.i18n.format( "TODO.ED.Chat.Flavor.Damage", {} )
+        chatFlavor: game.i18n.format( "TODO.ED.Chat.Flavor.Damage", {} ),
       },
       this._actor,
     );
@@ -160,7 +176,7 @@ export default class DamageWorkflow extends ActorWorkflow {
     const damageStep = await this.calculateDamageStep();
     return {
       step:       {
-        total:      damageStep ?? 0,
+        base:      damageStep ?? 0,
         modifiers: {},
       },
       target:     {


module/data/chat/attack.mjs
```diff
@@ -84,15 +84,30 @@ export default class AttackMessageData extends BaseMessageData {
...
+  //   event.preventDefault();
 
-    const weapon = await fromUuid( this.roll.options.weaponUuid );
-    if ( weapon?.system.roll instanceof Function ) return await weapon.system.rollDamage();
+  //   const weapon = await fromUuid( this.roll.options.weaponUuid );
+  //   if ( weapon?.system.roll instanceof Function ) return await weapon.system.rollDamage();
 
-    if ( this.roll.options.weaponType === "unarmed" ) return this.attacker.rollUnarmedDamage();
+  //   if ( this.roll.options.weaponType === "unarmed" ) return this.attacker.rollUnarmedDamage();
+  // }
```
ChristopherSD, line 94: Remove comments

module/data/chat/attack.mjs
```diff
@@ -84,15 +84,30 @@ export default class AttackMessageData extends BaseMessageData {
...
+  static async _onRolldamage( event, button ) {
```
ChristopherSD, line 95: Typo: camelCase!

module/data/chat/attack.mjs
```diff
@@ -11,7 +11,7 @@ export default class AttackMessageData extends BaseMessageData {
   static DEFAULT_OPTIONS = {
     actions: {
       "apply-effect":  this._onApplyEffect,
-      "roll-damage":   this._onRollDamage,
+      "roll-damage":   this._onRolldamage,
```
ChristopherSD, line 14: Typo: camelCase!

module/data/chat/attack.mjs
```diff
@@ -84,15 +84,30 @@ export default class AttackMessageData extends BaseMessageData {
...
-    if ( weapon?.system.roll instanceof Function ) return await weapon.system.rollDamage();
+  //   const weapon = await fromUuid( this.roll.options.weaponUuid );
+  //   if ( weapon?.system.roll instanceof Function ) return await weapon.system.rollDamage();
 
-    if ( this.roll.options.weaponType === "unarmed" ) return this.attacker.rollUnarmedDamage();
+  //   if ( this.roll.options.weaponType === "unarmed" ) return this.attacker.rollUnarmedDamage();
+  // }
+  static async _onRolldamage( event, button ) {
+    event.preventDefault();
+    let attackParameters = {
```
ChristopherSD, line 97: Why is it called _attack_Parameters when you hand it to _damage_ function, where the parameter is called _damageOptions_?

module/data/chat/attack.mjs
```diff
@@ -84,15 +84,30 @@ export default class AttackMessageData extends BaseMessageData {
...
+      weapon:     await fromUuid( this.roll.options.weaponUuid ) ?? null,
+      weaponType: this.roll.options.weaponType ?? null,
+      // powerType:  this.roll.options.powerType ?? null,
+      // spellType:  this.roll.options.spellType ?? null,
+      // armorType:  this...
```
ChristopherSD, line 102: Remove comments

module/documents/actor.mjs
```diff
@@ -753,6 +754,21 @@ export default class ActorEd extends Actor {
     };
   }
 
+  async damage ( damageOptions ) {
```
ChristopherSD, line 757: This needs to be documented. Especially the parameter and what is supposed to be in there.

module/documents/actor.mjs
```diff
@@ -753,6 +754,21 @@ export default class ActorEd extends Actor {
...
+    const damageWorkflow = new DamageWorkflow(
+      this,
+      {
+        armorType:      damageOptions.armorType,
+        damageItem:     damageOptions.weapon ?? null,
+        damageItemType: damageOptions.weaponType ?? null,
+        successes:      damageOptions.successes ?? 0,
+      },
```
ChristopherSD, line 765: This can all be solved with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
+import ActorWorkflow from "./actor-workflow.mjs";
+import RollPrompt from "../../applications/global/roll-prompt.mjs";
+import DamageRollOptions from "../../data/roll/damage.mjs";
+
+/**
+ * @typedef {object} DamageWorkflowOptions
+ * @property
```
ChristopherSD, line 7: If you don't have options, you don't need to define them

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+  _rollOptions;
+
+  /**
+   * The item being used for the damage.
+   * @type {ItemEd|undefined}
+   */
+  _damageItem;
+
+  /**
+   * the attribute used for the damage.
```
ChristopherSD, line 55: Typo: capitalize

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+      this.#rollDamage.bind( this ),
+    );
+  }
+
+  async #setDamageAttribute() {
+    // define the base attribute for the damage roll
+    if ( this._options.damageItem ) {
+      this._damageAttribute = this._actor.system.attributes[this._options.damageItem.system.damage.attribute].step ?? this._actor.system.attributes.str.step;
+    } else {
+      this._damageAttribute = this._actor.system.attributes.str;
```
ChristopherSD, line 87: There is a `.step` missing. Currently the object `str` is called

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+
+  // async #setSubstituteAbility() {
+  // }
+
+  // async #setExtraSuccesses() {
+  // }
+
+  async #setDamageType() {
+    // define the damage tpye for the damage roll
+    this._damageType = this._damageType ?? "leathal";
```
ChristopherSD, line 99: Why is this needed? It's not used anywhere?
Typo: "lethal"
Also, I think it should be "standard", which is what we use in the data model

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  async #setArmorType() {
+    // define the armor type for the damage roll
+    this._armorType = this._armorType ?? "physical";
```
ChristopherSD, line 104: This can be written as `this._armorType ??= "physical`
But I wonder why this is necessary, this property is never touched, so it can't have a value (it's only initialized without assignment, so it's always `undefined`)

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  async #setDamageItem() {
+    // define the damage item for the damage roll
+    this._damageItem = this._options.damageItem ?? this._actor.attributes.str.step;
```
ChristopherSD, line 109: Why not set it in the constructor?

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+      }
+    );
+  }
+
+  async #rollDamage() {
+    // currently used as intended, as roll processing is still part of the actor document
+    this._result = this._roll;
+  }
+
+  async calculateDamageStep() {
```
ChristopherSD, line 153: this is not a step or anything but is only used within this class and should therefore be private: add `#` before the name

module/workflows/workflow/damage-workflow.mjs
```diff
@@ -0,0 +1,198 @@
...
+        const totalWeaponStep = ( Number( this._damageAttribute ) + Number( this._damageItem.system.damage.baseStep ) );
+        damageStep += totalWeaponStep;
+      } else {
+        damageStep += this._damageAttribute;
+      };
+    }
+    return damageStep;
+  }
+
+  async getCommonDamageRollData() {
```
ChristopherSD, line 174: this is not a step or anything but is only used within this class and should therefore be private: add `#` before the name

ChristopherSD (changes requested):
This is currently an entirely different approach to damage than what we had before. We need to think and make some decisions on how we want to handel it. Is it an actor workflow? Or an item workflow (e.g. the weapon)?
Should it be one huge workflow for all different kinds, or multiple workflows for every type of damaging (could be one superclass basedamageworkflow that prepares the basic things, and the subclasses do the other stuff...)

PR #1660, opened 2025-04-29
remove isCreatureAttack ed-id

### LINKED ISSUE #1580: [REFACTORING]-remove ìscreatureAttack` from powerTemplate
remove ìscreatureAttack` from powerTemplate. it is handled via rolltypes

remove edId

patrickmohrmann:
resolve #1580

this will be handled via rolltype


ChristopherSD (changes requested):
Keep `isCreatureAttack` getter in PowerTemplate, but replace it with the new approach. Something like `` this.rollType === 'attack'`  maybe? Dunno...

ChristopherSD (approved):

PR #1664, opened 2025-05-03
create edid config

### LINKED ISSUE #947: Replace default value for edid settings with config values
Currently the default values for edIds in settings are string literals. They should be replaced with config values for more control and consistent use across the system.

patrickmohrmann:
resolve #947

### DIFF
--- a/module/config/system.mjs
+++ b/module/config/system.mjs
@@ -7,14 +7,35 @@ ________  _____       _____ _______
 |_______| |___/        |  | |______|
 _______________________________`;
 
+// region ED-IDs
+
 /**
  * Reserved earthdawn ids.
  * @enum {string}
  */
-export const reserved_edid = {
+export const reservedEdid = {
   DEFAULT:    "none",
   ANY:        "any",
 };
+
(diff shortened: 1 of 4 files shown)


ChristopherSD (commented):
.

module/config/system.mjs
```diff
@@ -69,4 +69,16 @@ export const icons = {
   },
 };
 
+export const edIds = {
```
ChristopherSD, line 72: I think we should call it something like "defaultEdids" to make very clear that they can be changed through settings.

ChristopherSD (changes requested):
Having the defaultNames in a config, I think it would make sense to create and register the settings dynamically from these values. Then the `getEdIds` function could also be made dynamic.

module/config/system.mjs
patrickmohrmann: I'm not 100% sure what you mean with a dynamic register of the edIds. lets talk about it together

ChristopherSD (approved):

PR #1665, opened 2025-05-04
fix issue that dropped talents have to show category prompt

### LINKED ISSUE #1658: [BUG]-Drag and drop talents onto a sheet is not asking for the category
Drag and drop talents onto a sheet is not asking for the category

patrickmohrmann:
resolve #1658


patrickmohrmann:
not needed

PR #1758, opened 2025-06-15
fix system settings options

patrickmohrmann:


module/hooks/sidebar.mjs
```diff
@@ -3,42 +3,66 @@ import RollPrompt from "../applications/global/roll-prompt.mjs";
  *
  */
 export default function () {
-  Hooks.on( "renderSidebarTab", ( app, html ) => {
-
+  Hooks.on( "renderSettings", ( app, html ) => {
     if ( app instanceof Settings ) {
```
ChristopherSD, line 7: Gives a deprecation warning now. Should be:
`if ( app instanceof foundry.applications.sidebar.tabs.Settings )`

module/hooks/sidebar.mjs
```diff
@@ -3,42 +3,66 @@ import RollPrompt from "../applications/global/roll-prompt.mjs";
...
-      } );
-      createBugButton.click( () => {
-        window.open( "https://github.com/patrickmohrmann/earthdawn4eV2/issues/new/choose", "_blank" );
-      } );
-    }
-  } );
+      const changeLogButton = document.createElement( "button" );
+      changeLogButton.id = "ed4eChangelog";
+      changeLogButton.className = "changelog";
+      changeLogButton.textContent = game.i18n.localize( "ED.Settings.SpecificSettingOptions.changelog" );
```
ChristopherSD, line 12: The localization should be a bit more informative. Specific, is, .. well, not very specific 😅
Also, none of these buttons are settings. They are more like, system related stuff. So maybe call it... "System Related Stuff", ED Stuff, or you know, something like that :D

ChristopherSD (changes requested):

module/hooks/sidebar.mjs
patrickmohrmann: I changed it to SystemSupportOptions, that might be a bit more fitting, but they are located in the system Settings area of foundry...

module/hooks/sidebar.mjs
patrickmohrmann: great.

ChristopherSD (approved):

Issue #656, opened 2024-04-05
roll prompt target defense


patrickmohrmann:
**Description**
the roll prompt titel shall show the target defense form the ability in the title.
e.g.: Melee weapons against targets physical defense.

on the defense tab it is possible to change that target defense, which will then update the title.

**Additional context**

PR #1789, opened 2025-07-01
ED Rolls Maintenance/Refactoring/Improvement

ChristopherSD:
Small overhaul of how ED Rolls work and it's prompt. Made several fields in RollOptions not required and nullable; `null` should indicate that the field is not used.
This is necessary for advanced roll types like warping rolls.
Additionally the roll prompt was freshened up, to have at least most options available. Also made more modular for later adaption.

### `RollPrompt` Class Refactor:

* Refactored `RollPrompt` to extend `ApplicationEd` instead of using `HandlebarsApplicationMixin`. This includes reorganizing methods into logical regions, adding static properties and methods, and improving rendering options. [[1]](diffhunk://#diff-78be9c83f55075556f1d061f31d5ece3084188594dc42f4a210b2238de068705R4-L9) [[2]](diffhunk://#diff-78be9c83f55075556f1d061f31d5ece3084188594dc42f4a210b2238de068705L25-R52) [[3]](diffhunk://#diff-78be9c83f55075556f1d061f31d5ece3084188594dc42f4a210b2238de068705L82-R160)
* Updated the `_configureRenderOptions` method to dynamically include parts like `step`, `target`, and `resources` based on the roll options.
* Reintroduced and reorganized methods for validating available resources and handling form submissions. [[1]](diffhunk://#diff-78be9c83f55075556f1d061f31d5ece3084188594dc42f4a210b2238de068705L137-R198) [[2]](diffhunk://#diff-78be9c83f55075556f1d061f31d5ece3084188594dc42f4a210b2238de068705R207-R229)

### Data Model Adjustments:

* Added a `defineSchema` method to the `SparseDataModel` class, returning an empty schema as a placeholder.
* Modified the default value of `step` in `EdRollOptions` from `1` to `0` and adjusted the `strain` field to be required but nullable. [[1]](diffhunk://#diff-13544520ffc9ba16c5f01ae045946e3a7aeaa0019385621460df6bd005757c76L105-R105) [[2]](diffhunk://#diff-13544520ffc9ba16c5f01ae045946e3a7aeaa0019385621460df6bd005757c76L181-R182)


lang/de.json
```diff
@@ -3200,6 +3224,7 @@
       },
       "Modifiers": {
         "grimoirePenalty":                         "Fremdes Grimoire",
+        "manual":                                  "Manuell",
```
patrickmohrmann, line 3227: I would suggest something like "Manuelle Stufenanpassung" or "Manuelle Anpassung"

lang/de.json
```diff
@@ -2554,15 +2578,15 @@
             "strain": {
               "base": {
                 "hint":                            "Basisbelastung des Wurfs",
-                "label":                           "Basisbelastung"
+                "label":                           "Überanstrengungsschaden - Basis"
```
patrickmohrmann, line 2581: I would suggest that "Überanstrengungsschaden" is enough, i don't see that "- Basis" is needed

templates/prompts/roll/part-step.hbs
```diff
@@ -0,0 +1,29 @@
+<fieldset>
+  <legend>{{localize "ED.Dialogs.RollPrompt.step"}}</legend>
+  {{formGroup
+    optionsFields.step.fields.base
+    value=step.base
+    hint=""
```
patrickmohrmann, line 6: shouldn't it be read only. we have the modifier field for adjustments and the Total test is beautifully showing the sum of it, so I think that should not be adjustable either

lang/de.json
```diff
@@ -2554,15 +2578,15 @@
...
               },
               "modifiers": {
                 "hint":                            "Modifikatoren die auf die Basisbelastung angewendet werden",
```
patrickmohrmann, line 2584: this is still "Basisbelastung" can you change it along with the others please?

templates/prompts/roll/part-target.hbs
```diff
@@ -0,0 +1,13 @@
+<fieldset>
+  <legend class="modifier-description">{{localize "ED.Dialogs.RollPrompt.difficulty"}}:</legend>
+  {{#if target.public}}
+    {{formGroup
+      optionsFields.target.fields.base
+      name="target.base"
```
patrickmohrmann, line 6: would be great if we could enhance the name "Basiszielwert" to something like "Verteidigung des Ziels" Ideally even adding the type of the defense to it

patrickmohrmann (changes requested):

lang/de.json
ChristopherSD: ok

lang/de.json
ChristopherSD: ok

templates/prompts/roll/part-target.hbs
ChristopherSD: Should make a new ticket for that?
I was thinking we're gonna improve on the roll prompt even more anyhow. Kind of have it like the roll, with different prompts (or rather, different `PARTS`) for the different roll types.

patrickmohrmann: updated #656 to cover this

templates/prompts/roll/part-step.hbs
patrickmohrmann: hm, my opinion is not that strong

patrickmohrmann (approved):

PR #1794, opened 2025-07-04
Rename pool options2

### LINKED ISSUE #578: Rename Pool options
The pool options for Paths and second thid and fourth dicsipline have different pool tiers

they should be configurable or auto configured 

paths start with journeyman up to master

questors have different names

patrickmohrmann:
resolves #578

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -609,6 +609,9 @@
         "summon":                                  "Beschwörung"
       },
       "Tier": {
+        "adherent":                                "Anhänger",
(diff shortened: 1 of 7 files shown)


ChristopherSD (commented):
This is pretty massive code in the template.
Could you just prepare all necessary things in the sheet when preparing the context?
so, basically, pre-filtered abilityOptions, so the smaller code still works?
That's also easier to debug...

PR #1806, opened 2025-07-08
1775 refactoring remove limitation for the first discipline to be only from char gen

### LINKED ISSUE #1775: [REFACTORING]-remove limitation for the first discipline to be only from char gen
its unhandy and in case of starting as a questor it is wrong to prohibit the drop for the first discipline outside of the char gen

patrickmohrmann:
resolves #1775

### DIFF
--- a/module/data/item/discipline.mjs
+++ b/module/data/item/discipline.mjs
@@ -4,7 +4,7 @@ import ED4E from "../../config/_module.mjs";
 import { linkForUuidSync } from "../../utils.mjs";
 import DescriptionMigration from "./migration/old-system-V082/description.mjs";
 
-const { expandObject, mergeObject } = foundry.utils;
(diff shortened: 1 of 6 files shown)


module/data/item/discipline.mjs
```diff
@@ -320,10 +320,6 @@ export default class DisciplineData extends ClassTemplate.mixin(
       ui.notifications.warn( game.i18n.localize( "ED.Notifications.Warn.cannotLearn" ) );
       return;
     }
-    if ( isEmpty( actor.disciplines ) ) {
-      ui.notifications.warn( game.i18n.localize( "ED.Notifications.Warn.firstDisciplineViaCharGen" ) );
```
patrickmohrmann, line 324: remove only return and adapt comment you need to use char gen, otherwise attributes are not working etc.

PR #1840, opened 2025-07-23
enable wound reduction if no recovery is actually rolled.

### LINKED ISSUE #1835: [BUG]-overnight rest not working with 0 damage
**Describe the bug**
if an actor has wounds and tries to take a full night rest, the wounds should be reduced by one, but instead nothing happens and the console. shows this error.

**log**
roll-processor.mjs:47 Uncaught (in promise) Error: RollProcessor.process: No roll provided
    at RollProcessor.process (roll-processor.mjs:47:24)
    at RecoveryWorkflow._processRecovery (recovery-workflow.mjs:250:25)
    at RecoveryWorkflow.execute (workflow.mjs:102:15)

patrickmohrmann:
resolves #1835

### DIFF
--- a/module/workflows/workflow/recovery-workflow.mjs
+++ b/module/workflows/workflow/recovery-workflow.mjs
@@ -247,8 +247,6 @@ export default class RecoveryWorkflow extends Rollable( ActorWorkflow ) {
       availableRecoveryTests -= 1;
     }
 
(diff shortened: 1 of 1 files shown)


ChristopherSD (changes requested):
I would sugest leaving the statement before the update, that should always be the last. Wrapping it in the `if` is enough.

ChristopherSD (approved):

PR #1841, opened 2025-07-23
Versatility talent limit validation

patrickmohrmann:
versatility talents now have a limitation as written in the rules

### DIFF
--- a/module/applications/global/prompt-factory.mjs
+++ b/module/applications/global/prompt-factory.mjs
@@ -444,11 +444,12 @@ class ActorPromptFactory extends PromptFactory {
 class ItemPromptFactory extends PromptFactory {
 
   _promptTypeMapping = {
-    chooseTier:     this._chooseTierPrompt.bind( this ),
-    learnKnack:     this._learnKnackPrompt.bind( this ),
-    lpIncrease:     this._lpIncreasePrompt.bind( this ),
-    learnAbility:   this._learnAbilityPrompt.bind( this ),
-    talentCategory: this._talentCategoryPrompt.bind( this ),
+    chooseTier:             this._chooseTierPrompt.bind( this ),
+    learnKnack:             this._learnKnackPrompt.bind( this ),
+    lpIncrease:             this._lpIncreasePrompt.bind( this ),
+    learnAbility:           this._learnAbilityPrompt.bind( this ),
+    talentCategory:         this._talentCategoryPrompt.bind( this ),
+    versatilityTalentLimit: this._versatilityTalentLimitPrompt.bind( this )
   };
 
   /**
@@ -642,6 +643,33 @@ class ItemPromptFactory extends PromptFactory {
       },
       modal:   false,
       buttons,
+    } );
+  }
+
+  /**
+   * Creates the versatility talent limit dialog.
+   * @returns {Promise<any>} A promise that resolves to the versatility talent limit prompt dialog
+   */
+  async _versatilityTalentLimitPrompt() {
+    const content = `<p>${ game.i18n.localize( "ED.Dialogs.versatilityTalentLimit" ) }</p>`;
+    
+    return DialogClass.wait( {
+      rejectClose: false,
+      id:          "talent-category-prompt",
+      uniqueId:    String( ++foundry.applications.api.ApplicationV2._appId ),
+      classes:     [ "earthdawn4e", "talent-category-prompt", "flexcol" ],
+      window:      {
+        title:       game.i18n.format( "ED.Dialogs.Title.versatilityLimit", {
+          abilityName: this.document.name,
+        } ),
+        minimizable: false
+      },
+      modal:   false,
+      content,
+      buttons: [
+        this.constructor.cancelButton,
+        this.constructor.continueButton
+      ]
     } );
   }
 }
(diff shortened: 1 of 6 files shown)


module/data/item/talent.mjs
```diff
@@ -243,6 +243,18 @@ export default class TalentData extends IncreasableAbilityTemplate.mixin(
       category = await promptFactoryItem.getPrompt( "talentCategory" );
     }
 
+    // versatility validation
+    if ( category === "versatility" ) {
+      const versatilityTalents = actor.items.filter(
```
ChristopherSD, line 248: Use `actor.itemTypes.talent.filter` to avoid the first condition.

module/data/item/talent.mjs
```diff
@@ -243,6 +243,18 @@ export default class TalentData extends IncreasableAbilityTemplate.mixin(
...
+        i => i.type === "talent" && i.system.talentCategory === "versatility"
+      );
+      const versatility = actor.items.filter( item => item.type === "talent" && item.system.edid === "versatility" );
```
ChristopherSD, line 251: We have a method for that on the actor document:
`this.containingActor?.getSingleItemByEdid(...)`

module/data/item/talent.mjs
```diff
@@ -243,6 +243,18 @@ export default class TalentData extends IncreasableAbilityTemplate.mixin(
...
+      if ( versatilityTalents.length === versatility[0].system.level ) {
+        ui.notifications.warn( game.i18n.localize( "ED.Notifications.Warn.versatilityTalentLimit" ) );
+        return;
```
ChristopherSD, line 254: I would not be so strict to just not allow it. Rather prompt them with the information and ask if they want to continue anyways. 
That's the philosophy, right? Automate everything, Dictate (almost) nothing...

module/data/item/talent.mjs
```diff
@@ -259,7 +271,7 @@ export default class TalentData extends IncreasableAbilityTemplate.mixin(
     // update the learned talent with the new data
     await learnedItem.update( {
       "system.talentCategory":        category ?? learnedItem.system.talentCategory,
-      "system.source.class":          learnedItem.system.source?.class ?? discipline.uuid,
+      "system.source.class":          learnedItem.system.source?.class ?? discipline?.uuid,
```
ChristopherSD, line 274: This way, this field can be `undefined`, should that be possible when learning something with this function?

ChristopherSD (changes requested):

module/data/item/talent.mjs
patrickmohrmann: great

module/data/item/talent.mjs
patrickmohrmann: good idea to just throw a warning, taht should be enough

module/data/item/talent.mjs
patrickmohrmann: yes that's exactly necessary, Otherwise you are not able to "just add a talent because of xyz" I think this is rather necessary in terms of not forbidding anything

ChristopherSD: hmm... well... yeah, I guess? as long as it doesn't fuck up any of our other workflows. There's a limitation, coming from the data model. But if that works, yeah. You can resolve this comment if you want

module/data/item/talent.mjs
ChristopherSD: Well, actually. You could just prompt them, like with the LP spending. Ask them if the want to continue anyways? If not, rules obeyed, if yes, they do their own thing...

patrickmohrmann: oh thats a very good idea, way better then just a warning 👍

module/data/item/talent.mjs
```diff
@@ -243,6 +243,17 @@ export default class TalentData extends IncreasableAbilityTemplate.mixin(
       category = await promptFactoryItem.getPrompt( "talentCategory" );
     }
 
+    // versatility validation
+    if ( category === "versatility" ) {
+      const versatilityTalents = actor.itemTypes.talent.filter(
+        item =>  item.system.talentCategory === "versatility"
+      );
+      const versatility = this.containingActor?.getSingleItemByEdid( "versatility" );
+      if ( versatilityTalents.length === versatility[0].system.level ) {
```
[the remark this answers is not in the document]
patrickmohrmann: ok

module/applications/global/prompt-factory.mjs
```diff
@@ -644,4 +645,31 @@ class ItemPromptFactory extends PromptFactory {
...
+        minimizable: false
+      },
+      modal:   false,
+      content,
+      buttons: [
+        this.constructor.cancelButton,
+        this.constructor.continueButton
+      ]
+    } );
+  }
```
ChristopherSD, line 674: This can be simplified with `DialogClass.confirm(...)` 
No need for buttons, classes for styling, too options, etc. You can think about doing it in the code over there directly, bc this prompt is only relevant in this one spot. So this function will only ever be called once, so it's not really necessary. And the confirm method is short enough to do it there (or in a small function over there)

ChristopherSD (changes requested):

module/applications/global/prompt-factory.mjs
patrickmohrmann: I didn't know about this. nice

PR #2259, opened 2025-08-06
768 new actor type npc

### LINKED ISSUE #768: New Actor Type NPC
with the Actor Type Character done, the next in line is the NPCs, which is very similar to the Character with the option to have every field manually editable instead of automatic calculated.

Npcs shall be very similar to characters, with two major differences:
1. calculated characteristic fields can be overwritten easily. The values shall all be set individual. 
2. Equipments do not add bonuses.
3. Effects add bonuses
4. the option to use attacks, powers and maneuvers (the later only for tracking reasons, without function)
Things to be aware off:
- Tab order of the sheet values
- Each tab shall automatically select everything inside of the field
- Npc Token shall be neutral by default and have sight

Everything else is like the pc

patrickmohrmann:
resolve #768 

some minor changes to the NPC actor to make it workable
+ default token settings

### DIFF
--- a/module/config/actors.mjs
+++ b/module/config/actors.mjs
@@ -103,14 +103,3 @@ export const languages = {
   windling:     "ED.Languages.windling",
 };
 preLocalize( "languages" );
-
-/**
- * Token Dispositions
- * @enum {number}
- */
-export const tokenDisposition = {
-  friendly:   CONST.TOKEN_DISPOSITIONS.FRIENDLY,
-  neutral:    CONST.TOKEN_DISPOSITIONS.NEUTRAL,
-  hostile:    CONST.TOKEN_DISPOSITIONS.HOSTILE,
-  secret:     CONST.TOKEN_DISPOSITIONS.SECRET,
-};
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -128,31 +128,8 @@ export default class ActorEd extends Actor {
   /** @inheritDoc */
   async _preCreate( data, options, userId ) {
     await super._preCreate( data, options, userId );
-    await this._updateTokenSettings();
-  }
-
-  async _updateTokenSettings () {
-    const actorsTypesWithKarma = [ "character", "npc" ];
-    const prototypeToken = {
-      sight:       {enabled: true},
-      actorLink:   this.type === "character",
-      disposition: await this._getDisposition(),
-      displayBars: 50,  // Always Display bar 1 and 2
-      displayName: 30,  // Display nameplate on hover
-      bar1:        {
-        attribute: "healthRate"
-      },
-      bar2: {
-        attribute: actorsTypesWithKarma.includes( this.type ) ? "karma" : null
-      }
-    };
-    this.updateSource( { prototypeToken } );
-  }
-
-  async _getDisposition() {
-    if ( this.type === "character" ) return ED4E.tokenDisposition.friendly;
-    if ( this.type === "npc" ) return ED4E.tokenDisposition.neutral;
-    return ED4E.tokenDisposition.hostile;
+    const prototypeToken = ED4E.prototypeToken[this.type];
+    await this.updateSource( { prototypeToken } );
   }
 
   /**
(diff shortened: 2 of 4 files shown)


module/documents/actor.mjs
```diff
@@ -128,25 +128,31 @@ export default class ActorEd extends Actor {
...
-        displayBars: 50,  // Always Display bar 1 and 2
-        displayName: 30,  // Display nameplate on hover
-        bar1:        {
-          attribute: "healthRate"
-        },
-        bar2: {
-          attribute: "karma"
-        }
-      };
+  async _tokenSettingUpdate () {
```
ChristopherSD, line 134: It's best practice to name functions as "doing something", so, at least to me, it feels more natural to read something like `updateTokenSetting` or sth like that

module/documents/actor.mjs
```diff
@@ -128,25 +128,31 @@ export default class ActorEd extends Actor {
...
+    const actorsTypesWithKarma = [ "character", "npc" ];
```
ChristopherSD, line 135: Not for now, but we really need to make config enum for our document types. We already got so many magic strings in the code with that * starts sweating slightly * :D

module/documents/actor.mjs
```diff
@@ -128,25 +128,31 @@ export default class ActorEd extends Actor {
...
+    const prototypeToken = {
+      sight:       {enabled: true},
+      actorLink:   this.type === "character" ? true : false,
```
ChristopherSD, line 138: The ternary is redundant here and the same as `this.type === "character"` since that evaluates to a boolean anyway

module/documents/actor.mjs
```diff
@@ -128,25 +128,31 @@ export default class ActorEd extends Actor {
...
+    this.updateSource( { prototypeToken } );
+  }
 
-      this.updateSource( { prototypeToken } );
-    }
+  async _getDisposition() {
+    if ( this.type === "character" ) return CONST.TOKEN_DISPOSITIONS.FRIENDLY;
+    if ( this.type === "npc" ) return CONST.TOKEN_DISPOSITIONS.NEUTRAL;
+    return CONST.TOKEN_DISPOSITIONS.UNFRIENDLY;
   }
```
ChristopherSD, line 156: Good idea. Since that is really only a lookup, I would put that in a config enum, easier to manage, and maybe useful for later functionality (like attitudes...)

ChristopherSD (changes requested):

module/documents/actor.mjs
patrickmohrmann: if its not for now, i think thats great, because i have no clue what you refering to :-P

module/documents/actor.mjs
patrickmohrmann: ah I always forget that this works like this

module/documents/actor.mjs
```diff
@@ -128,25 +128,31 @@ export default class ActorEd extends Actor {
...
+    this.updateSource( { prototypeToken } );
+  }
 
-      this.updateSource( { prototypeToken } );
-    }
+  async _getDisposition() {
+    if ( this.type === "character" ) return ED4E.tokenDisposition.friendly;
+    if ( this.type === "npc" ) return ED4E.tokenDisposition.neutral;
+    return ED4E.tokenDisposition.hostile;
   }
```
ChristopherSD, line 156: That is not what I meant... Rather, the config should have what's written here, so you don't need the function, but rather just can write this up top:
`disposition: ED4E.tokenDisposition[this.type]`

module/config/actors.mjs
```diff
@@ -102,4 +102,15 @@ export const languages = {
...
+/**
+ * Token Dispositions
+ * @enum {number}
+ */
+export const tokenDisposition = {
+  friendly:   CONST.TOKEN_DISPOSITIONS.FRIENDLY,
+  neutral:    CONST.TOKEN_DISPOSITIONS.NEUTRAL,
+  hostile:    CONST.TOKEN_DISPOSITIONS.HOSTILE,
+  secret:     CONST.TOKEN_DISPOSITIONS.SECRET,
+};
```
ChristopherSD, line 116: That's not what I meant. This is just doubling what's already there in `CONST.TOKEN_DISPOSITIONS`. I meant something like:
```ecmascript 6
export const defaultTokenDisposition = {
  character: CONST.TOKEN_DISPOSITIONS.FRIENDLY,
  npc:          CONST.TOKEN_DISPOSITIONS.NEUTRAL,
  //...
};
```

ChristopherSD (changes requested):

ChristopherSD (approved):

PR #2280, opened 2025-08-07
fix partial migration

### LINKED ISSUE #2279: [BUG]-
Migration has issues:

updating status of an equipment is running migration each time, why is the protection not working?

reattuning a matrix stumbles accross abilities line 11 slugify because it has no name...

patrickmohrmann:
migration has to be protected from updateEmbeddedDocuments, which does Partial updates.
resolves #2279

### DIFF
--- a/module/documents/actor.mjs
+++ b/module/documents/actor.mjs
@@ -1117,6 +1117,16 @@ export default class ActorEd extends Actor {
 
   // region Migrations
   static migrateData( source ) {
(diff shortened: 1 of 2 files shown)


ChristopherSD (changes requested):
This is repeated/doubled code and should therefore be refactored into a function. Thtat could be part of the "migrations" module that is still to come/ is in `services` right now.

ChristopherSD (approved):

PR #2371, opened 2025-08-24
Item to macro favorites

### LINKED ISSUE #1532: [NEW FEATURE]- Show favorites on the first tab
~abilities and all rollable items in general shall be able to be set as "favorite" those items shall show up on the general tab with only the option to roll them and to edit (for reading purposes).~

instead of having an option to select an item as favorite, a context menu is used to via right click. If the context menu is chosen, two things happen:
- a macro is created for this item
  - the macro has to be created for the user (inside an folder with the user as owner)
- the macro is added to the favorite area of the actor sheet.

### LINKED ISSUE #2351: delete favorites
# Story
there shall be an option to delete favorites from the general tab of the actor sheet.

- the macro shall not be deleted automatically

# Detailed Description
(optional)

# Acceptance Criteria
**Given** [context]  
**When** [event]  
**Then** [expected outcome]

### LINKED ISSUE #2355: enable spells to have the favorite context menu
# Story
spell cards shall have the context menu option to become a favorite (and create a macro etc)

# Detailed Description
(optional)

# Acceptance Criteria
**Given** [context]  
**When** [event]  
**Then** [expected outcome]

patrickmohrmann:
big one

folder creation is missing, its already big

resolves #1532 
resolves #2351 
resolves #2355

### DIFF
--- a/lang/pl.json
+++ b/lang/pl.json
@@ -3066,6 +3066,7 @@
         }
       },
       "DeleteFavorite": {
+        "alsoDeletesMacro":                        "TODO: Add translation",
         "favorite":                                "TODO: Add translation"
       },
       "Header": {
@@ -3087,13 +3088,13 @@
           "SelectedSpells":                        "TODO: Add translation",
           "YouCanChooseSpells":                    "TODO: Add translation",
           "chooseAbilityOption":                   "TODO: Add translation",
+          "classAbilitiesGained":                  "TODO: Add translation",
           "classIncreasedTier":                    "TODO: Add translation",
           "effectsGained":                         "TODO: Add translation",
           "increaseClassTo":                       "TODO: Add translation",
           "learningClass":                         "TODO: Add translation",
           "optionalAbilityGained":                 "TODO: Add translation",
           "resourceStepIncreased":                 "TODO: Add translation",
-          "classAbilitiesGained":                  "TODO: Add translation",
           "spellsGained":                          "TODO: Add translation"
         },
         "LearnSpell": {
@@ -3369,6 +3370,7 @@
     },
     "Notifications": {
       "Error": {
+        "abilityIncreaseNoTier":                   "TODO: Add translation",
         "actorWarningSingleton":                   "TODO: Add translation",
         "formulaMalformedError":                   "TODO: Add translation",
         "formulaMissingReferenceWarn":             "TODO: Add translation",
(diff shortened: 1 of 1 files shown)


module/applications/actor/common-sheet.mjs
```diff
@@ -102,6 +139,58 @@ export default class ActorSheetEd extends DocumentSheetMixinEd( ActorSheetV2 ) {
...
+          "system.favorites": updatedFavorites
+        } );
+      },
+      options: {
+        top:   Math.min( event.clientY - 80, window.innerHeight - 350 ),
+        left:  window.innerWidth - 720,
+        width: 400
+      }
+    } );
+  }
```
ChristopherSD, line 178: This currently only removes the macro from favorites, but the document still exists. Are you sure you want to do that? That could really clutter the macro collection.
Maybe rather get the macro document and do `deleteDialog()`? And then if there's a document returned, that means yes and we can remove it from favorites?

module/data/item/armor.mjs
```diff
@@ -88,4 +88,15 @@ export default class ArmorData extends PhysicalItemTemplate.mixin(
       } ),
     } );
   }
+
+  // region Macros
+
+  /** @inheritDoc */
+  getDefaultMacroCommand( options = {} ) {
+    console.log( "ArmorData.getDefaultMacroCommand called for:", this.parent.name );
```
ChristopherSD, line 96: please remove console logging like this

module/documents/item.mjs
```diff
@@ -157,6 +157,30 @@ export default class ItemEd extends Item {
...
+    }
+
+    const macroData = {
+      name:       this.name,
+      type:       CONST.MACRO_TYPES.SCRIPT,
+      img:        this.img,
+      command:    this.system.getDefaultMacroCommand(),
+    };
+
+    // TODO: Add macro to folder if it exists, add user specific folder if not
```
ChristopherSD, line 179: This should be part of this PR. I don't really see a reason to do it later...

ChristopherSD (changes requested):
There's a lot of copyPaste in the item data models for the default macro command. That should be changed. If it's about rollable items, put it into rollable. If its about physical items that can be rolled, put it into PhysicalItems.

module/applications/actor/common-sheet.mjs
patrickmohrmann: i tried it and found it to be a good option, you can still keep the macro if you want to, but I would say, lets talk about it, last time that was very good and the result was awesome too

module/documents/item.mjs
patrickmohrmann: ok, then i need help, folders can only be created as GM not as user

module/applications/actor/common-sheet.mjs
ChristopherSD: We decided to add an additional context entry to "create macro". This also creates a macro document but doesn't add it to the favorites. When deleting a favorite, the corresponding document will be deleted as well.
See #2531

PR #2379, opened 2025-08-26
fix bug

patrickmohrmann:
fixes potential issue, that at some point old value old promption becomes an object before the source is updated. so the next time, old values are objects instead of numbers


module/services/migrations/v082/document-type-migrations/actor/character.mjs
```diff
@@ -67,8 +67,16 @@ export default class CharacterMigration extends BaseMigration {
       // Only migrate if we have old data and the new format doesn't already exist
       if ( oldValue !== undefined && !source.system.attributes[newName]?.hasOwnProperty( "initialValue" ) ) {
         source.system.attributes[newName] ??= {};
-        source.system.attributes[newName].initialValue = oldValue;
-        source.system.attributes[newName].timesIncreased = oldPromotions ?? 0;
+        if ( typeof oldValue === "object" && oldValue.hasOwnProperty( "initialValue" ) ) {
```
ChristopherSD, line 70: Maybe use https://foundryvtt.com/api/functions/foundry.utils.getType.html . Could be a bit better and more thorough?

ChristopherSD (approved):

PR #2380, opened 2025-08-26
bring back take damage button

patrickmohrmann:
readded the take damage button,


alegionofone (approved):
Added as button, based on an svg image, all other buttons on the page have been styled to be the same (gold-4 with gold-5 when hovering). This functions over all NPCs and the character sheet

PR #2381, opened 2025-08-26
ensure indices are documents

### LINKED ISSUE #1860: [BUG]-Skills in compendium are not accounted for char gen
[BUG]-Skills in compendium are not accounted for char gen

patrickmohrmann:
the symple indices might just by stripped down information while documents are full documents. predicate expect documents

resolves #1860

### DIFF
--- a/module/utils.mjs
+++ b/module/utils.mjs
@@ -163,7 +163,18 @@ export async function getAllDocuments(
     } ),
   ).then( p => p.flat() );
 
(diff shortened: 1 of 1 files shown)


ChristopherSD (commented):
Okay. Then we might need to think about it a little more. If we change this function, or if we change the char generation.
getting all the documents from all indices is costly since it's a database operation. So, whenever this function is called, _all_ documents that exist with the predicate are fetched, that can be potentially hundreds...
Let's go through this and a resolution together

PR #2473, opened 2025-08-30
En sheet fixes

alegionofone:


lang/fr.json
```diff
@@ -1993,8 +2002,12 @@
...
               "label":                             "TODO: Add translation"
             },
+            "damageType": {
+              "hint":                              "TODO: Add translation",
+              "label":                             "TODO: Add translation"
+            },
             "element": {
-              "subtype": {
+              "subType": {
                 "hint":                            "TODO: Add translation",
```
patrickmohrmann, line 2011: did you change this in the french lang file on purpose, it should be subtype not subType I guess

lang/fr.json
```diff
@@ -891,6 +891,15 @@
             "label":                               "TODO: Add translation"
           }
         },
+        "Group": {
```
patrickmohrmann, line 894: don't readd them please

patrickmohrmann (changes requested):

PR #2486, opened 2025-09-02
fix bug

### LINKED ISSUE #1478: [BUG]-Editing Legend History loses navigation tab active
editing and saving the Legend Point history, will remove the **active** class from the actor sheet navigation.

patrickmohrmann:
resolve #1478

add a default render method to remember the last active tab


ChristopherSD:
We have to recheck all of our Tab handling after using the V13 built-in functionality now...

PR #2490, opened 2025-09-04
remove none existiing activateTab function

### LINKED ISSUE #1809: [BUG]-char generation creates a bug
Error message on char generation 

pc.mjs:187 Uncaught TypeError: actorApp.activateTab is not a function
    at pc.mjs:187:43

patrickmohrmann:
fix #1809

### DIFF
--- a/module/data/actor/pc.mjs
+++ b/module/data/actor/pc.mjs
@@ -202,9 +202,7 @@ export default class PcData extends NamegiverTemplate {
       }
     }
 
-    const actorApp = newActor.sheet.render( true, {focus: true} );
-    // we have to wait until the app is rendered to activate a tab
-    requestAnimationFrame( () => actorApp.activateTab( "actor-notes-tab" ) );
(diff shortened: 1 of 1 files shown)


ChristopherSD (approved):

module/data/actor/pc.mjs
```diff
@@ -202,9 +202,7 @@ export default class PcData extends NamegiverTemplate {
       }
     }
 
-    const actorApp = newActor.sheet.render( true, {focus: true} );
-    // we have to wait until the app is rendered to activate a tab
-    requestAnimationFrame( () => actorApp.activateTab( "actor-notes-tab" ) );
+    newActor.sheet.render( true, {focus: true} );
```
ChristopherSD, line 205: this is an async method and should have an `await`

ChristopherSD (changes requested):

module/data/actor/pc.mjs
patrickmohrmann: ok

PR #2494, opened 2025-09-05
1722 lp tracking include entitytype

### LINKED ISSUE #1722: LP Tracking include entityType
lp-spending-transaction shall use the entityType to enable backwards tracking

patrickmohrmann:
resolves #1722 

basically clean up and data provisioning for localization in all sortings

### DIFF
--- a/module/data/advancement/lp-tracking.mjs
+++ b/module/data/advancement/lp-tracking.mjs
@@ -118,8 +118,6 @@ export default class LpTrackingData extends foundry.abstract.DataModel {
 
   get htmlChronological() {
     const rows = this.chronologically.map( transaction => {
-      transaction.name ??= "";
-      
       return `
       <tr class="" data-group="" data-id="${ transaction.id }">
           <td>
@@ -129,7 +127,7 @@ export default class LpTrackingData extends foundry.abstract.DataModel {
             ${ game.i18n.localize( "ED.Dialogs.Legend.LpHistory." + transaction.type ) }
           </td>
           <td>
-            ${ transaction.name }
+            ${ transaction.name ?? "" }
           </td>
           <td>
             ${ transaction.description }


module/data/advancement/lp-spending-transaction.mjs
```diff
@@ -70,6 +70,16 @@ export default class LpSpendingTransactionData extends LpTransactionData {
...
   getHtmlRow( index, classes, dataGroup ) {
+    // Generate a user-friendly description if the current one is a localization key
+    let displayDescription = this.description;
+    if ( displayDescription === "ED.Actor.LpTracking.Spendings" && this.value?.before !== undefined && this.value?.after !== undefined ) {
+      // Generate a description similar to what learnable items use
+      displayDescription = game.i18n.format( "ED.Actor.LpTracking.Spendings.spendingTransactionDescription", {
+        previousLevel: this.value.before,
+        newLevel:      this.value.after
+      } );
+    }
```
ChristopherSD, line 81: I'm not sure this is necessary. It should be up to the code that creates this data model to provide a useful description. Which it usually does. This is why for example the `LpIncreaseTemplate` has the `lpSpendingDescription` property getter.
In what cases was this description not user-friendly? This is probably the point to start and fix, I would assume...

module/data/advancement/lp-tracking.mjs
```diff
@@ -118,13 +118,19 @@ export default class LpTrackingData extends foundry.abstract.DataModel {
 
   get htmlChronological() {
     const rows = this.chronologically.map( transaction => {
+      // Determine if this transaction has a name (spending transactions have names)
+      const itemName = transaction.name ? transaction.name : "";
```
ChristopherSD, line 122: This constant is not needed, you can just put that expression down there, without the comment. Also, it can be written as: `transaction.name ?? """`

module/data/advancement/lp-tracking.mjs
```diff
@@ -233,10 +242,14 @@ export default class LpTrackingData extends foundry.abstract.DataModel {
     const content = Object.values( groupedData ).map(
       group => {
         const category = group[0]?.category;
+        // Show raw category name when sorting by "item", but localized when sorting by "type"
+        const displayCategory = categoryType === "name" 
+          ? category 
+          : game.i18n.localize( "ED.Dialogs.Legend.LpHistory." + category );
```
ChristopherSD, line 248: This feels to specific. What if another category is added at some point? It would be split up into even more conditions.
I would suggest, trying to make it more flexible to accomodate any kind of category.
And why can't name fit into the earlier approach? Can the name not be localized? If there is no localize key, it just takes the input string anyway.

ChristopherSD (changes requested):
The new display description gives me only `null`s:
image
See comment in lp-spending-transactions.mjs

The display category doesn't change anything for me. There is no `categoryType` "name".

module/data/advancement/lp-spending-transaction.mjs
patrickmohrmann: yes thats right, i reverted that

module/data/advancement/lp-tracking.mjs
ChristopherSD: now you're assigning a name to the transaction, that's not what I meant. And that definitely shouldn't be done in a rendering function.
just add `transaction.name ?? "` down in the curly braces. This line is not necessary

patrickmohrmann: ah ok, i was using the wrong operator, i got it now

PR #2518, opened 2025-09-14
add tier to language skills

### LINKED ISSUE #1882: [BUG]-Char gen creates read and speak skills over and over again
starting the charakter generation will create everytime a new pair of speak language and read/write language skill, even though they are availabel. this might very well be linked to the bug #1860

patrickmohrmann:
without the tier the skills are not found in the collection

resolves #1882


ChristopherSD (changes requested):
I do appreciate trying not to use magic strings, but :D
In this case, you can actually just write "novice", since those config objects are Enumerations. The problem with magic xyz is that they are not explained/cannot be understood by themselves. In this case, we have no idea what the elemnt at index 0 is. So just write "novice" :)

PR #2648, opened 2025-10-05
Add true pattern to physical (thread) items

ChristopherSD:
Physical items now have a true pattern instead of threads.
This is the first step towards non-spell thread magic.
Thread handling comes in the next PR.

To add a true pattern to an item, the sheet must be in edit mode.
The "create/weave thread" button doesn't have a function yet.


module/applications/item/physical-item-sheet.mjs
```diff
@@ -156,20 +160,50 @@ export default class PhysicalItemSheetEd extends ItemSheetEd {
...
+    await this.render();
+  }
+
+  /**
+   * @type {ApplicationClickAction}
+   * @this {PhysicalItemSheetEd}
+   */
+  static async _onDeleteThreadItemLevel( event, target ) {
+    event.preventDefault();
+    await this.document.system.truePattern.removeLastThreadItemLevel();
```
patrickmohrmann, line 187: there should be a delete question as well? I would go for the quick delete option as well here?

module/data/thread/thread-item-level.mjs
```diff
@@ -0,0 +1,61 @@
...
+          nullable: false,
+          initial:  "",
+        } ),
+        isKnown: new fields.BooleanField( {
+          required: true,
+          nullable: false,
+          initial:  false,
+        } )
+      } ),
+      deed: new fields.StringField( {
```
patrickmohrmann, line 39: no deedComplete flag? I know its not needed really, but would be nice to have a check there

module/data/thread/thread-item-level.mjs
```diff
@@ -0,0 +1,61 @@
...
+        required: true,
+        nullable: false,
+        initial:  "",
+      } ),
+      effect: new fields.StringField( {
+        required: true,
+        nullable: false,
+        initial:  "",
+      } ),
+      activeEffects: new fields.SetField(
```
patrickmohrmann, line 49: I thought about adding an additional pool for "everything else". some items like the Butter-Spider-Box have "special Abilities" which are more than just an EAE

ChristopherSD: Added `abilities` to add any number of items (like special abilities for example)

PR #2669, opened 2025-10-10
add castingRank as rolltype

patrickmohrmann:
simple additional rolltype option to use for active effects


patrickmohrmann:
broken

PR #2748, opened 2025-10-17
✨ Weaving threads

qrizzl:
Functionality to weave new threads to thread items, from the thread item itself (button weave thread). This creates a new item of type "thread" on the chosen actor. Actor is chosen in this order, whichever matches first: 
1.  containing actor of the thread item
2. the assigned character of the user
3. the first controlled token
4. one chosen from a prompt of all characters

Increasing a thread rank also means weaving a thread. In Code, this means increasing the rank, but with the entire thread weaving workflow beforehand.

Add WeaveThreadWorkflow to validate requirements and make thread weaving test. Currently only considers thread items for requirements.

Weaving roll options had to be adapted to both cover spell thread weaving and general thread weaving, Better test for correct spell weaving too, I guess?

### DIFF
--- a/module/data/thread/true-pattern.mjs
+++ b/module/data/thread/true-pattern.mjs
@@ -222,6 +222,18 @@ export default class TruePatternData extends SparseDataModel {
     return LEGEND.legendPointsCost[ newLevel + tierModifier ];
   }
 
+  /**
+   * Get the amount of legend points required to increase the entity to the given level.
+   * @param {number} [level] The level to get the required legend points for. Defaults to the next level.
+   * @returns {number|undefined} The amount of legend points required to increase the entity to the given
+   * level. Or `undefined` if the amount cannot be determined.
+   */
+  getRequiredLpForLevelSync( level ) {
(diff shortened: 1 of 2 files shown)


patrickmohrmann:
LP calculation incorrect:
image

module/data/thread/true-pattern.mjs
```diff
@@ -173,6 +173,22 @@ export default class TruePatternData extends SparseDataModel {
...
+  get numberOfAttachedThreads() {
+    return this.attachedThreads?.size ?? 0;
+  }
+
+  /**
+   * Whether more threads can be attached to this true pattern.
+   * @type {boolean}
+   */
+  get canHaveMoreThreads() {
+    return this.numberOfAttachedThreads < this.maxThreads;
```
patrickmohrmann: canHaveMoreThreads is called when i do thread weaving for level 2 or three from the item itself. Is it by intention?

I think one actor should only be able to weave one thread and if it is already woven it shall either increase the thread or be disabled don't you think? (is fine as a refactoring for later though)

patrickmohrmann (approved):

PR #2821, opened 2025-10-26
second attempt for spellcasting rolltypes

patrickmohrmann:


module/data/item/spell.mjs
```diff
@@ -373,7 +389,11 @@ export default class SpellData extends ItemDataModel.mixin(
   getRollData() {
     const rollData = super.getRollData();
     Object.assign( rollData, super.getTemplatesRollData() );
-    return Object.assign( rollData, {} );
+    return Object.assign( rollData, {
+      castingRank:      this.getSpellcastingRank(),
+      spellcastingRank: this.getSpellcastingRank(),
+      rank:             this.getSpellcastingRank(),
+    } );
```
ChristopherSD: Here you handle the undefined case then, which also lets you call the function only once, which is more efficient. Something like:
```javascript
const spellcastingRank = this.getSpellcastingRank() ?? 0;
return Object.assign( rollData, {
    castingRank:         spellcastingRank,
    spellcastingRank: spellcastingRank,
    rank:                     spellcastingRank,
  } );
```

ChristopherSD (changes requested):

module/data/item/spell.mjs
```diff
@@ -326,6 +326,22 @@ export default class SpellData extends ItemDataModel.mixin(
...
+  getSpellcastingRank( ) {
+    let spellcastingRank = 0;
+    const spellcastingTalent = this.containingActor?.getSingleItemByEdid(
+      game.settings.get( "ed4e", "edidSpellcasting" ),
+      "talent",
+    );
+    if ( spellcastingTalent ) {
+      spellcastingRank = spellcastingTalent.system.level;
+    }
+    return spellcastingRank;
```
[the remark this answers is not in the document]
patrickmohrmann: aha, that makes perfect sense

module/data/item/spell.mjs
patrickmohrmann: ah its not about the if its not there then use 0 its about the different functions, yes yes, got it

PR #3274, opened 2025-12-08
🐛 Fix Talent Rank Cost in Multi Discipline

### LINKED ISSUE #3173: [BUG]- Multi discipline initial talent rank cost calculation not correct
**Describe the bug**
the increased cost calculation for the first rank of the discipline talents of a second, third or forth discipline are not correctly calculated. 

weaponssmith C5 learning wizard showed on one instance 1300 lp as rank 1 cost (should be 200)
thief C5 learning illusionist showed on one instance 0 for rank 1 but second rank was ok

**finding** it looks like if the new discipline is counted as lowest discipline for the increase calculation. if i manually increase that second discipline, the required LP is reduced accordingly.

qrizzl:
Fix #3173

### DIFF
--- a/lang/pl.json
+++ b/lang/pl.json
@@ -3174,10 +3174,12 @@
           "classAbilitiesGained":                  "TODO: Add translation",
           "classIncreasedTier":                    "TODO: Add translation",
           "effectsGained":                         "TODO: Add translation",
+          "freeAbilitiesGained":                   "TODO: Add translation",
           "increaseClassTo":                       "TODO: Add translation",
           "learningClass":                         "TODO: Add translation",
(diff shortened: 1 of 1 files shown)


ChristopherSD:
in `TalentData`'s `get requiredLpForIncrease` don't use the `minCircle` getter but make a `getMinCircle` function where you can ignore a given discipline.

Decisions:
- do not show "Required LP" in class advancement prompts
- do not auto increase added talents when adding another discipline (only "learn" discipline and optional talents at level 0), manually increase them to one each (maybe add an info prompt for that?)

patrickmohrmann (approved):

PR #3434, opened 2026-01-02
Fix powers to be rollable

patrickmohrmann:
this enables powers to be rolled from the actual tab 

in addition, i enabled the EAEs extended card views to be activatable

### DIFF
--- a/templates/actor/cards/power-card.hbs
+++ b/templates/actor/cards/power-card.hbs
@@ -1,10 +1,9 @@
 
 <div class="card__power favoritable" data-uuid="{{uuid}}">
     <div class="power-card__grid--container item-id" data-item-id="{{_id}}">
-        {{log "this" this.name this.system.rollType}}
-        <div class="power-card__grid--item icons__dice {{#if (or (ne this.system.powerStep 0) (ne this.system.strain 0) (ne this.system.rollType ""))}}has-overlay{{/if}}">
+        <div class="power-card__grid--item icons__dice {{#if (and (or (ne this.system.powerStep 0) (ne this.system.strain 0)) this.system.rollType)}}has-overlay{{/if}}">
         <img class="power-icon" src="{{img}}" alt="{{name}}">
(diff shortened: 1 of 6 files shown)


templates/actor/cards/power-card.hbs
```diff
@@ -1,26 +1,38 @@
...
+                type="button"
+                class="icon--button icon icon--d20 rollable icon--overlay"
+                data-action="rollable"
+                data-roll-type="ability"
+                data-tooltip="Roll {{name}}"
+                data-tooltip-direction="UP"
+                aria-label="Roll {{name}}">
+            </button>
+        {{else}}
+            {{#if (and (ne this.system.strain 0) this.system.rollType)}}
```
ChristopherSD: Why do I need a rollType if it doesn't have a step and therefore not a roll?

templates/global/effect-card.hbs
```diff
@@ -20,8 +20,8 @@
     <div class="effect-card__grid--item">{{> "systems/ed4e/templates/global/card-options-effect.hbs" }}</div>
   </div>
   <div class="card__description card__description--toggle display__inline-block">
-    <span class="card__description--content">
-      {{{system.description.value}}}
+    <span class="card__description--content">}
```
ChristopherSD: there's a stray curly brace at the end of the line. needs to be deleted, otherwise it will be rendered like that

patrickmohrmann: oh good catch

patrickmohrmann:
the idea is simple. (and yes the styling was not perfect).

almost all creatures have powers which are not really rolled but provide effects. Some of them are easy, because they are always "active" but others like frenzy, auras, manifestations, etc. (there are a lot) need to be activated only when needed. 

you mentioned that those effects should not be handled on the actor sheet. The alternative would be 3 clicks more for every time plus an extra item dialog to close again, that is a lot for a gm. 

but I can see the thrash bin to be out of place for these effects.

PR #3461, opened 2026-01-24
🐛 3056 bug substitute attacks use highest rank for all

### LINKED ISSUE #3056: [BUG]-Substitute attacks use highest rank for all
**Describe the bug**
if an actor has a talent e.g. Melee weapons and uses a substitute attack of Missile Weapons for example, the rank of the melee weapons talent is used

**To Reproduce**
1. have only one attack ability with at least 1 rank
2. make an attack with a substitute skill unlike the attack ability
3. the rolled step is equal to the attack ability not dexterity

**Expected behavior**
the substitute shall only have the bonus if the combat type (melee, unarmed, etc.) is equal

**Additional context**
its unlikely, but it might be worth referencing ed-ids or something like that to cover edge cases like an actor having second weapon but no melee weapons

### LINKED ISSUE #3182: [BUG] ensure off hand weapons do not trigger weapon choice prompt
# Story
ensure that the choose weapon prompt only appears for off hand attacks if there is no off hand equipped.

# Detailed Description
(optional)

# Acceptance Criteria
**Given** [context]  
**When** [event]  
**Then** [expected outcome]

patrickmohrmann:
Fixes #3056 
Fixes #3182 

Substitute attacks are now working and not taken always the highest attack rank 

in addition, moved switch weapon, draw weapon and check equipped weapon to actor.mjs from ability.mjs

### DIFF
--- a/module/data/item/templates/ability.mjs
+++ b/module/data/item/templates/ability.mjs
@@ -270,24 +270,16 @@ export default class AbilityTemplate extends ActionTemplate.mixin(
   async rollAttack() {
     if ( !this.isActorEmbedded ) return;
 
-    const whatToDo = await this.containingActor.checkEquippedWeapons( this );
-    if ( !whatToDo ) throw new Error( "No action to take! Something's messed up :)" );
+    const weaponData = await this.containingActor.prepareWeaponForAttack( this );
+    if ( !weaponData ) return;
 
-    let weapon = null;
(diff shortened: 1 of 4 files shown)


qrizzl (commented):
Updated a few things, please re-check if it still works.
This now also includes #3182 since it is part of the refactored method.

qrizzl (approved):

PR #3653, opened 2026-03-25
🚩Add feature flag for selectively creating document system types

### LINKED ISSUE #2755: [REFACTORING]-remove or hide not (yet) supported items and actors
some items might not be required and some might not be needed yet or have not enough functionality.

same for actors

those shall be removed or hidden

qrizzl:
Resolves #2755

IMPORTANT!
Look at `UNAVAILABLE_SYSTEM_TYPES` and decide which ones should be included.

Any system types written there cannot be created via the UI.

### DIFF
--- a/module/config/system.mjs
+++ b/module/config/system.mjs
@@ -141,8 +141,8 @@ export const FLAGS = {
  * @type {{Actor: string[], Item: string[]}}
  */
 export const UNAVAILABLE_SYSTEM_TYPES = {
-  Actor: [ "dragon", "horror", "loot", "spirit", "trap", "vehicle" ],
-  Item:  [ "bindingSecret", "curseMark", "mask", "poisonDisease", "shipWeapon", "spellKnack", ],
+  Actor: [ "loot", "trap", "vehicle" ],
+  Item:  [ "bindingSecret", "curseMark", "poisonDisease", "shipWeapon", ],
 };
 
 // endregion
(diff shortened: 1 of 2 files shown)


module/config/system.mjs
```diff
@@ -136,4 +136,13 @@ export const FLAGS = {
   learnedSpellKnackThroughPath: "learnedSpellKnackThroughPath",
 };
 
+/**
+ * The system types that are not available in the current system and should not be shown when creating a new document.
+ * @type {{Actor: string[], Item: string[]}}
+ */
+export const UNAVAILABLE_SYSTEM_TYPES = {
+  Actor: [ "dragon", "horror", "loot", "spirit", "trap", "vehicle" ],
```
patrickmohrmann: only, loot, vehicle, trap and group should not be used currently

module/config/system.mjs
```diff
@@ -136,4 +136,13 @@ export const FLAGS = {
...
+  Item:  [ "bindingSecret", "curseMark", "mask", "poisonDisease", "shipWeapon", "spellKnack", ],
```
patrickmohrmann: spellKnacks and Masks are used, the rest is ok

patrickmohrmann (changes requested):

module/config/system.mjs
qrizzl: Don't we need "group" for things like true pattern/thread magic?

patrickmohrmann (approved):

PR #3687, opened 2026-04-08
♻️👽️🏗️Update Active Effect system for Foundry v14 compatibility

qrizzl:
# This is breaking changes! Only for Foundry v14 upwards

Update Earthdawn Active Effects to Foundry V14.

### Checklist
- [x] Check EAE on actors
- [x] Check EAE on items
- [x] Check correct calculation of properties for actors (e.g., death rating)
- [x] Check correct calculation and auto-update of class effects (like discipline+questor+path)
- [x] Check correct trigger of EAE scripts


patrickmohrmann:
Findings:

- Assign Effects - Not working
- Effects for **other** actors - not working, always on the source
- removing all change entries from an effect and then clicking + will throw an error
image

individual findings per attribute key (Actor effects):
- we have two times spellcasing (all spellcasting and all spellcasting tests
- durability bonus is not working
- Blood wounds are not added to wounds and do not reduce the rolls (not sure if this ever worked)

individual findings per attribute key (Item effects for Items):
- Difficulty --> not working 
- fixed difficulty --> not working
- attribute --> no effect (does this one make sense?
- tier --> no effect (does this one make sense?

some more findings:
- Dropping Effects into path/questor item throws an error
- effects from level up are inactive by default
- effects from the same discipline do not cancle each other out if they are higher. they stack.

qrizzl:
> Findings:
> 
>     * Assign Effects - Not working
> 
>     * Effects for **other** actors - not working, always on the source
> 
>     * removing all change entries from an effect and then clicking + will throw an error
> 
> 
> 
> individual findings per attribute key (Actor effects):
> 
>     * we have two times spellcasing (all spellcasting and all spellcasting tests
> 
>     * durability bonus is not working
> 
>     * Blood wounds are not added to wounds and do not reduce the rolls (not sure if this ever worked)
> 
> 
> individual findings per attribute key (Item effects for Items):
> 
>     * Difficulty --> not working
> 
>     * fixed difficulty --> not working
> 
>     * attribute --> no effect (does this one make sense?
> 
>     * tier --> no effect (does this one make sense?
> 
> 
> some more findings:
> 
>     * Dropping Effects into path/questor item throws an error
> 
>     * effects from level up are inactive by default
> 
>     * effects from the same discipline do not cancle each other out if they are higher. they stack.

- assign effects never worked -> not part of this PR
- I dont know what you mean with "other actors"
- adding change when non exist is fixed

- two times spellcasting: explicitly separated between "spellcasting" (the talent use) and "spell tests" (including effect rolls for example) -> not part of this PR
- durability bonus works now
- blood wounds never worked -> not part of this PR

-  which "Difficulty"? there is no single "difficulty" key I think
- "fixed difficulty" works for me
- "attribute" is a string and should be changed by using "override" and then add the value, like "str" for Strength
- same for "tier"

- dropping effects on path/questor items works for me 
- effects from level up is fixed
- class effect updates is fixed

patrickmohrmann (approved):

PR #3720, opened 2026-04-13
:bug: :recycle: removed min restriction to masks

### LINKED ISSUE #3719: [BUG] - Mask input values cannot be negative
### Bug Checklist

### What Happened?

Mask input values cannot be negative

they should

### Steps to Reproduce

add -1

v13

FoundryVTT Desktop App

### Anything Else?

_No response_

patrickmohrmann:
the min: 0 data model restriction was wrong for most fields

### DIFF
--- a/module/data/item/mask.mjs
+++ b/module/data/item/mask.mjs
@@ -55,7 +55,7 @@ export default class MaskData extends ItemDataModel.mixin(
     return this.mergeSchema( super.defineSchema(), {
       attributes: new MappingField( new fields.SchemaField( {
         step: new fields.NumberField( {
-          min:      0,
+          required: true,
           initial:  0,
           integer:  true,
         } )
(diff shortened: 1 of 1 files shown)


module/data/item/mask.mjs
```diff
@@ -66,34 +66,38 @@ export default class MaskData extends ItemDataModel.mixin(
...
           nullable: true,
           integer:  true,
         } ),
         climb: new fields.NumberField( {
           required: true,
+          initial:  0,
           nullable: true,
           integer:  true,
         } )
       } ),
```
qrizzl, line 92: If these are initially 0 and are expected to be a number, then they shouldn't be `nullable`, so I'd remove these (if they are `false` by default, otherwise write it down explicitly)

qrizzl (changes requested):

qrizzl (approved):

PR #3860, opened 2026-05-21
🐛 Open Earthdawn Active Effects from Effect Cards

### LINKED ISSUE #3858: [BUG] - open EAE not working
### Bug Checklist

### What Happened?

opening the "Elf" namegiver from the namegiver players guide compendium throwns an error:

Uncaught (in promise) TypeError: Cannot read properties of null (reading 'sheet')
    at ItemSheetEd._onEditChild (document-sheet-mixin.mjs:264:49)
_onEditChild @ document-sheet-mixin.mjs:264
await in _onEditChild
#onClickAction @ foundry.mjs:31784
#onClick @ foundry.mjs:31744

effect would open

### Steps to Reproduce

1. activate players guide
2. open elf from the namegiver compendium
3. try to open the eae "low-light Vision"

FVTT 14

Chrome, FoundryVTT Desktop App

### Anything Else?

_No response_

patrickmohrmann:
Resolve #3858 

add item id as identifier for own effects which do not have a uuid

### DIFF
--- a/module/applications/api/document-sheet-mixin.mjs
+++ b/module/applications/api/document-sheet-mixin.mjs
@@ -161,6 +161,35 @@ const DocumentSheetMixinEd = Base => {
     // region Event Handlers
 
     /**
+     * Falls back to nearest row data-item-id when no UUID is available in template context.
+     * @param {HTMLElement} target - The HTML element that triggered the event.
+     * @returns {Promise<foundry.abstract.Document|null>} - The resolved child document, or null if not found.
+     * @protected
+     */
+    async _resolveChildDocument( target ) {
+      const uuid
+        = target.dataset.uuid
+        ?? target.dataset.effectUuid
+        ?? target.closest( "[data-uuid]" )?.dataset.uuid;
+
+      if ( uuid ) {
+        const document = await fromUuid( uuid );
+        if ( document ) return document;
+      }
+
+      const embeddedId
+        = target.dataset.itemId
+        ?? target.dataset.effectId
+        ?? target.closest( "[data-item-id]" )?.dataset.itemId;
+
+      if ( !embeddedId ) return null;
+
+      return this.document.items?.get( embeddedId )
+        ?? this.document.effects?.get( embeddedId )
+        ?? null;
+    }
+
+    /**
      * Handle the user toggling the sheet mode.
      * @param {Event} event  The triggering event.
      * @protected
@@ -239,9 +268,13 @@ const DocumentSheetMixinEd = Base => {
      * @returns {Promise<foundry.abstract.Document>} - A promise that resolves when the child is deleted.
      */
     static async _onDeleteChild( event, target ) {
-      const document = await fromUuid( target.dataset.uuid );
+      const document = await this._resolveChildDocument( target );
+      if ( !document ) return;
+
       if ( getSetting( "quickDeleteEmbeddedOnShiftClick" ) && event.shiftKey ) return document.delete();
-      else document.deleteDialog();
+
+      if ( typeof document.deleteDialog === "function" ) return document.deleteDialog();
+      return document.delete();
     }
 
     /**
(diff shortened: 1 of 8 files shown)


module/applications/api/document-sheet-mixin.mjs
```diff
@@ -160,6 +160,35 @@ const DocumentSheetMixinEd = Base => {
...
+
+      if ( uuid ) {
+        const document = await fromUuid( uuid );
+        if ( document ) return document;
+      }
+
+      const embeddedId
+        = target.dataset.itemId
+        ?? target.dataset.effectId
+        ?? target.closest( "[data-item-id]" )?.dataset.itemId;
```
qrizzl: This is dangerous i think, because it could grab the id of another unrelated item and then that one gets deleted.
Remove this and let it be undefined.

module/applications/api/document-sheet-mixin.mjs
```diff
@@ -160,6 +160,35 @@ const DocumentSheetMixinEd = Base => {
...
+     * Falls back to nearest row data-item-id when no UUID is available in template context.
+     * @param {HTMLElement} target - The HTML element that triggered the event.
+     * @returns {Promise<foundry.abstract.Document|null>} - The resolved child document, or null if not found.
+     * @protected
+     */
+    async _resolveChildDocument( target ) {
+      const uuid
+        = target.dataset.uuid
+        ?? target.dataset.effectUuid
+        ?? target.closest( "[data-uuid]" )?.dataset.uuid;
```
qrizzl: This is dangerous i think, because it could grab the id of another unrelated item and then that one gets deleted. Remove this and let it be undefined.

module/applications/api/document-sheet-mixin.mjs
```diff
@@ -160,6 +160,35 @@ const DocumentSheetMixinEd = Base => {
...
+    /**
+     * Falls back to nearest row data-item-id when no UUID is available in template context.
+     * @param {HTMLElement} target - The HTML element that triggered the event.
+     * @returns {Promise<foundry.abstract.Document|null>} - The resolved child document, or null if not found.
+     * @protected
+     */
+    async _resolveChildDocument( target ) {
+      const uuid
+        = target.dataset.uuid
+        ?? target.dataset.effectUuid
```
qrizzl: There's no need for "effectUuid". The whole idea of UUIDs is that it identifies everything about the resource, including its document type.

module/applications/api/document-sheet-mixin.mjs
```diff
@@ -160,6 +160,35 @@ const DocumentSheetMixinEd = Base => {
...
+        ?? target.closest( "[data-uuid]" )?.dataset.uuid;
+
+      if ( uuid ) {
+        const document = await fromUuid( uuid );
+        if ( document ) return document;
+      }
+
+      const embeddedId
+        = target.dataset.itemId
+        ?? target.dataset.effectId
```
qrizzl: In the effect cards you used "itemId", so why do we need this here? Either change it there or here. I'd prefer to keep it generic, which would mean finding a generic word, like `document-id`, and refactor _all_ templates that include this.

module/applications/api/document-sheet-mixin.mjs
```diff
@@ -160,6 +160,35 @@ const DocumentSheetMixinEd = Base => {
...
+        ?? target.closest( "[data-item-id]" )?.dataset.itemId;
+
+      if ( !embeddedId ) return null;
+
+      return this.document.items?.get( embeddedId )
+        ?? this.document.effects?.get( embeddedId )
+        ?? null;
```
qrizzl: What happens if for some reasons we have both an `Item` and an `ActiveEffect` with the same ID?

qrizzl (changes requested):

module/applications/api/document-sheet-mixin.mjs
patrickmohrmann: i changed it to have the data-uuid as primary lookup

module/applications/api/document-sheet-mixin.mjs
patrickmohrmann: ok, i adapted the templates as well

module/applications/api/document-sheet-mixin.mjs
patrickmohrmann: ok

module/applications/api/document-sheet-mixin.mjs
patrickmohrmann: ok i removed it

module/applications/api/document-sheet-mixin.mjs
patrickmohrmann: ok i removed it

qrizzl (approved):

PR #3884, opened 2026-05-29
💄 Make Roll Prompt Step Modifiers Collapsible

qrizzl:
Improve information overload by making step modifiers collapsible.

This is just a suggestion because I heard some feedback that a more simple prompt would help.
The big overhaul and nice-making of the roll prompt of course is still waiting.

### DIFF
--- a/templates/prompts/roll/part-step.hbs
+++ b/templates/prompts/roll/part-step.hbs
@@ -13,14 +13,16 @@
     disabled=true
   }}
   {{#if ( ne testType "arbitrary") }}
-    <label class="modifier-description">{{localize "ED.Dialogs.RollPrompt.modifier"}}</label>
-    {{#each step.modifiers as |modifier|}}
-      {{formGroup
-        ../optionsFields.step.fields.modifiers.element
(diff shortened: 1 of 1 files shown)


patrickmohrmann:
ok, persönlich finde ich es nicht schön, ich muss jetzt jedes mal auf den pfeil klicken weil ich wissen will ob alles stimmt, wenn wir noch ein client setting einbauen wäre ich einverstanden

qrizzl:
@patrickmohrmann do you alway check all modifiers for correctness? I'm assuming most of the users (i.e., players) will not always check. that's why it would make the prompt easier on the eyes and "mind" as well...
A setting is not really viable I think (the purpose of a `summary` block is to be closed, because the rest is detail that's not always needed).
I was just making assumptions about the broader user base. We don't have to do it now, feel free to deny the PR :)

patrickmohrmann:
I get the point, and at some point it might be viable, but at the moment I'm really always checking

qrizzl:
Will be done later

PR #3891, opened 2026-06-12
🐛 Don't overwrite already existing properties of Spell Knack Data

### LINKED ISSUE #3869: [BUG] - spellknack description is overwritten by Source ED-ID spell
### Bug Checklist

### What Happened?

when i enter an ed id of a spell into the ed-id source field of a spellknack, the description of the source ED-ID spell is taken over into the spellknacks description.

this should not happen

### Steps to Reproduce

create a spellknack
add a description
add the spell ed id

V14 

FoundryVTT Desktop App

### Anything Else?

_No response_

patrickmohrmann:
Fixes #3869

### DIFF
--- a/module/data/item/spell-knacks.mjs
+++ b/module/data/item/spell-knacks.mjs
@@ -96,9 +96,12 @@ export default class SpellKnackData extends SpellData.mixin(
       : await getSingleGlobalItemByEdid( data.system.sourceItem, SYSTEM_TYPES.Item.spell );
     if ( !sourceSpell ) return;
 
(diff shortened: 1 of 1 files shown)


qrizzl (changes requested):
If you wanna only fill data that doesn't exist yet I would rather choose a more universal approach. This currently takes care of "description", what if, at some point, some other stuff gets overwritten?
So I'd recommend changing the `overwrite` in `mergeObject` to `false`. This way, only missing properties are added.

qrizzl (approved):

PR #3946, opened 2026-07-22
🐛 Make active effect source origin editable

patrickmohrmann:
somehow this change got lost. it just enables the update of the effect origin


qrizzl (approved):
I feel like we had some reason for it, but I can't remember, so 🤷

